PR #136, opened 2025-11-25
Update BT23_030.cs

ghost:
This one is a shot in the dark. Updating the Main effect to add cardSource.IsTamer to the hand selection targets.


x89rt (commented):
makes sense

PR #155, opened 2025-11-26
Update P_117.cs

ghost:
This is a wild shot. Code taken from BT20-046 Espimon, changed trait to appropriate, added tamer condition, added a hashstring, and prayed.

### DIFF
--- a/CardEffect/P/Blue/P_117.cs
+++ b/CardEffect/P/Blue/P_117.cs
@@ -19,7 +19,7 @@ namespace DCGO.CardEffects.P
             {
                 ActivateClass activateClass = new ActivateClass();
                 activateClass.SetUpICardEffect("Reduce the digivolution cost by 1", CanUseCondition, card);
-                activateClass.SetUpActivateClass(CanActivateCondition, ActivateCoroutine, -1, false, EffectDiscription());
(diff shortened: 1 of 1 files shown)


x89rt:
Hm... I think I see the issue. If I use the Gato method, it will show the discount on hover but it won't have the OPT. If I use the old Espi method, it should have the OPT but not the discount on hover.... which is probably preferable for now. Your call.

CardEffect/P/Blue/P_117.cs
```diff
@@ -15,136 +15,154 @@ public override List<ICardEffect> CardEffects(EffectTiming timing, CardSource ca
...
-            costReduceClass.SetHashString("DigivolutionCost-1_P_117");
-
-            string EffectDiscription2()
+            if (timing == EffectTiming.BeforePayCost)
             {
-                return "[Your Turn] [Once Per Turn] When this Digimon would digivolve into a Digimon card with the [Free] trait, if you have a Tamer, reduce the digivolution cost by 1.";
-            }
+                ActivateClass activateClass = new ActivateClass();
+                activateClass.SetUpICardEffect("Reduce the digivolution cost by 1", CanUseCondition, card);
+                activateClass.SetUpActivateClass(CanActivateCondition, ActivateCoroutine, -1, false, EffectDiscription());
```
Michael8818: we can try this change, im uncertain about it. Though it for sure need the -1 changed to 1 for the OPT

PR #198, opened 2025-11-26
Update BT19_049.cs

ghost:
Taking a shot out of removing the needing an opponent's digimon to trigger.


x89rt (commented):
Used bt17-026 as base to hopefully do this one right.

PR #471, opened 2025-11-29
Create ST22-03.cs

ghost:
I did it, but I'm a little worried that it's checking if it can do OM and WD at the trigger timing instead of at the activation timing.


x89rt (commented):
updated the text description to be more accurate

PR #472, opened 2025-11-29
Create ST22_04.cs

ghost:
I think I got most of it. Outside of the commented out code that I'm working on to see if I can figure out the use an option from under the tamer bit confusing my bracketing, it should be fine. A hell of a Frankenstein made from like 5 different cards.

### DIFF
--- a/CardEffect/ST22_04.cs
+++ b/CardEffect/ST22_04.cs
@@ -3,6 +3,7 @@ using System.Collections;
 using System.Collections.Generic;
 using System.Linq;
 
+// Taomon
 namespace DCGO.CardEffects.ST22
 {
 	public class ST22_04 : CEntity_Effect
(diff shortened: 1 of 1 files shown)


CardEffect/ST22_04.cs
```diff
@@ -0,0 +1,457 @@
...
+                        }
+                        
+                        if (CardEffectCommons.IsPermanentExistsOnOwnerBattleAreaTamer(permanent, card))
+                        {
+                            totalSourceCount = permanent.DigivolutionCards.Count(CanSelectCardCondition);
+                            if (totalSourceCount >= 1)
+                            {
+                                return true;
+                            }
+                        }
```
Michael8818: use these instead:

CardEffectCommons.HasMatchConditionOwnersHand(CanSelectCardCondition)

CardEffectCommons.HasMatchConditionPermanent(CanSelectTamerCondition)

They return true/false if you have a card in hand/ a permanent on field that meets the conditions passed, in this case the proper option in hand/in sources of a tamer

PR #649, opened 2025-12-05
Implement P-225

### LINKED ISSUE #646: P-225 - BT24 Box Topper
Image

hooperk:
Closes #646

### DIFF
--- a/CardEffect/P/White/P_225.cs
+++ b/CardEffect/P/White/P_225.cs
@@ -1,0 +1,189 @@
+using System;
+using System.Collections;
+using System.Collections.Generic;
+
+// DigiLab
+namespace DCGO.CardEffects.P
+{
+    public class P_225 : CEntity_Effect
+    {
+        public override List<ICardEffect> CardEffects(EffectTiming timing, CardSource card)
+        {
+            List<ICardEffect> cardEffects = new List<ICardEffect>();
+
+            #region Ignore Color Requirment
+
+            if (timing == EffectTiming.None)
+            {
+                IgnoreColorConditionClass ignoreColorConditionClass = new IgnoreColorConditionClass();
+                ignoreColorConditionClass.SetUpICardEffect("Ignore color requirements", CanUseCondition, card);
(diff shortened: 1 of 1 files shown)


CardEffect/P/White/P_225.cs
```diff
@@ -0,0 +1,185 @@
...
+                            SelectPermanentEffect selectPermanentEffect = GManager.instance.GetComponent<SelectPermanentEffect>();
+                            int maxCount = Math.Min(1, CardEffectCommons.MatchConditionPermanentCount(card, CanSelectCard));
+
+                            selectPermanentEffect.SetUp(
+                                selectPlayer: card.Owner,
+                                canTargetCondition: CanSelectCard,
+                                canTargetCondition_ByPreSelecetedList: null,
+                                canEndSelectCondition: null,
+                                maxCount: maxCount,
+                                canNoSelect: false,
```
x89rt2: Correct me if I'm wrong, but the canNoSelect should be true, no? You can pop the option, choose not to do the BY, not get the memory, and have it go to trash. Might also require a verification on if you did the thing to gain the mem or not.

CardEffect/P/White/P_225.cs
```diff
@@ -0,0 +1,185 @@
...
+                bool CanActivateCondition(Hashtable hashtable)
+                {
+                    return CardEffectCommons.IsExistOnBattleArea(card);
+                }
+
+                bool CanSelectCard(Permanent permanent)
+                {
+                    return permanent.IsDigimon
+                        && permanent.HasCSTraits
+                        && permanent.level >= 4
```
Michael8818, line 118: for extra safety add a check for permanent.TopCard.hasLevel

CardEffect/P/White/P_225.cs
Michael8818: this is correct it should be true

Michael8818 (changes requested):

Michael8818 (approved):

PR #651, opened 2025-12-06
Implement ST22-08

### LINKED ISSUE #488: ST22-08
Image

BennAbbot:
Feature: https://github.com/DCGO2/DCGO-Card-Scripts/issues/488

Instead of using regions like the other files, I separate the effects into functions. I find this cleaner, easier to read and looking at some of the effects that share triggers like `[On Play][When Digivolving]` would reduce the duplicate code I have seen. 

However this is isn't my project so if it is required to follow the format that the other effect class use I can change it.

This card has a lot going on so I expect something to be missing.

Also I have no idea if the second part of the [Main] effect is correct in letting chose you a low dp digimon to avoid deleting something the wording they used for the effect seems to be new so cant find any older cards to look at for reference.

### DIFF
--- a/CardEffect/ST22/Red/ST22_08.cs
+++ b/CardEffect/ST22/Red/ST22_08.cs
@@ -1,0 +1,313 @@
+using System;
+using System.Collections;
+using System.Collections.Generic;
+
+//ST22 Offensive Plug-in V
+namespace DCGO.CardEffects.ST22
+{
+    public class ST22_08 : CEntity_Effect
+    {
+        public override List<ICardEffect> CardEffects(EffectTiming timing, CardSource card)
+        {
+            List<ICardEffect> cardEffects = new List<ICardEffect>();
+
+            switch (timing)
+            {
+                case EffectTiming.None:
+                    cardEffects.Add(IgnoreColorCondition(card));
+                    cardEffects.Add(LinkCondtion(card));
+                    break;
+                case EffectTiming.OnDeclaration:
+                    cardEffects.Add(LinkAction(card));
+                    break;
+                case EffectTiming.OnEndTurn:
+                    cardEffects.Add(EndOfTurnLinkedEffect(card));
+                    break;
+                case EffectTiming.SecuritySkill:
+                    cardEffects.Add(SecurityEffect(card));
+                    break;
+                case EffectTiming.OptionSkill:
+                    cardEffects.Add(MainEffect(card));
+                    break;
+
+            }
+
+            return cardEffects;
+        }
+
+        IgnoreColorConditionClass IgnoreColorCondition(CardSource card)
+        {
+            IgnoreColorConditionClass ignoreColorConditionClass = new IgnoreColorConditionClass();
+            ignoreColorConditionClass.SetUpICardEffect("Ignore color requirements", CanUseCondition, card);
+            ignoreColorConditionClass.SetUpIgnoreColorConditionClass(cardCondition: CardCondition);
+
+            bool CanUseCondition(Hashtable hashtable)
+            {
+                return CardEffectCommons.HasMatchConditionOwnersPermanent(card, (permanent) => permanent.IsTamer);
+            }
+
+            bool CardCondition(CardSource cardSource)
+            {
+                return cardSource == card;
+            }
+
+            return ignoreColorConditionClass;
+        }
+
+        ActivateClass SecurityEffect(CardSource card)
+        {
+            ActivateClass activateClass = new ActivateClass();
+            activateClass.SetUpICardEffect($"Delete opponent's lowest dp Digimon and add this card to hand", CanUseCondition, card);
+            activateClass.SetUpActivateClass(null, ActivateCoroutine, -1, false, EffectDiscription());
+            activateClass.SetIsSecurityEffect(true);
+
+            string EffectDiscription()
+                => "[Security] Delete 1 of your opponent's Digimon with the lowest DP. Then, add this card to the hand.";
+           
+            bool CanUseCondition(Hashtable hashtable)
+            {
+                return CardEffectCommons.CanTriggerSecurityEffect(hashtable, card);
+            }
+
+            bool CanSelectPermanentCondition(Permanent permanent)
+                => CardEffectCommons.IsMinDP(permanent, card.Owner.Enemy);
+
+            IEnumerator ActivateCoroutine(Hashtable _hashtable)
+            {
+                int maxCount = Math.Min(1, CardEffectCommons.MatchConditionPermanentCount(CanSelectPermanentCondition));
+
+                SelectPermanentEffect selectPermanentEffect = GManager.instance.GetComponent<SelectPermanentEffect>();
+
+                selectPermanentEffect.SetUp(
+                    selectPlayer: card.Owner,
+                    canTargetCondition: CanSelectPermanentCondition,
+                    canTargetCondition_ByPreSelecetedList: null,
+                    canEndSelectCondition: null,
+                    maxCount: maxCount,
+                    canNoSelect: false,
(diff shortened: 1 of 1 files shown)


CardEffect/ST22/Red/ST22_08.cs
```diff
@@ -0,0 +1,313 @@
...
+                yield return ContinuousController.instance.StartCoroutine(CardEffectCommons.AddThisCardToHand(card, activateClass));
+            }
+
+            return activateClass;
+        }
+
+        ActivateClass MainEffect(CardSource card)
+        {
+            ActivateClass activateClass = new ActivateClass();
+            activateClass.SetUpICardEffect(card.BaseENGCardNameFromEntity, CanUseCondition, card);
```
BennAbbot, line 105: Also does the SetUpICardEffect's `effectname` get shown to the players at all if its not an optional effect? I copied this line from one of the old plugins, while other examples use a description of the effect.

CardEffect/ST22/Red/ST22_08.cs
```diff
@@ -0,0 +1,313 @@
...
+                => "[Main] You may link this card to 1 of your Digimon without paying the cost. Then, delete 1 of your opponent's Digimon with as much or less DP as 1 of your Digimon.";
+
+            bool CanUseCondition(Hashtable hashtable)
+                => CardEffectCommons.CanTriggerOptionMainEffect(hashtable, card);
+
+            IEnumerator ActivateCoroutine(Hashtable _hashtable)
+            {
+                #region Select Digimon To Link
+                {
+                    bool CanSelectPermanentCondition(Permanent permanent)
```
x89rt2: Feels like it needs to check that the digimon has levels and that the level is 3 or higher here.

CardEffect/ST22/Red/ST22_08.cs
```diff
@@ -0,0 +1,313 @@
...
+
+            return activateClass;
+        }
+
+        ActivateClass LinkAction(CardSource card)
+        {
+            return CardEffectFactory.LinkEffect(card);
+        }
+
+        AddLinkConditionClass LinkCondtion(CardSource card)
```
x89rt2: condition is misspelled

x89rt2:
The DP comparison and some of the other things went a little over my head so take my comments with a grain of salt.

CardEffect/ST22/Red/ST22_08.cs
```diff
@@ -0,0 +1,313 @@
...
+                            return false;
+                        }
+
+                        return CardEffectCommons.HasMatchConditionOpponentsPermanent(card, (opponentPermanent) =>
+                        {
+                            return opponentPermanent.IsDigimon && opponentPermanent.DP <= ownerPermanent.DP;
+                        });
+                    });
+
+                    bool CanSelectPermanentCondition(Permanent permanent)
```
LewisWelch94: You already have a function with this name line 117-121
`
{
    return CardEffectCommons.IsPermanentExistsOnOwnerBattleAreaDigimon(permanent, card) &&
            card.CanLinkToTargetPermanent(permanent, false);
}
`

I would rename apporiately

CardEffect/ST22/Red/ST22_08.cs
```diff
@@ -0,0 +1,313 @@
...
+                        }
+                    }
+                }
+                #endregion
+
+                Permanent comparePermanent = null;
+
+                #region Select Digimon to Compare
+                {
+                    bool opponentHasAnyTarget = CardEffectCommons.HasMatchConditionOwnersPermanent(card, (ownerPermanent) =>
```
LewisWelch94: This should be within your boolean function

LewisWelch94:
Suggestion regarding the Main 2nd effect.

`
#region Delete Digimon Setup

Permanent ownerSelectDigimon = null;

List<(Permanent MyDigimon, Permanent OpponentDigimon)> matches = new();

bool ValidDigimon()
{
    var myDigimon = card.Owner.GetBattleAreaDigimons();
    var opponentDigimon = card.Owner.Enemy.GetBattleAreaDigimons();

    matches = myDigimon
        .SelectMany(my =>
            opponentDigimon
                .Where(opp => opp.DP <= my.DP)
                .Select(opp => (My: my, Opponent: opp))
        )
        .ToList();

    return matches.Any();
}

bool CanSelectOwnerDigimon(Permanent permanent)
{
    return CardEffectCommons.IsPermanentExistsOnOwnerBattleAreaDigimon(permanent, card)
        && matches.Exists(x => x.MyDigimon == permanent);
}

bool CanSelectOpponentDigimon(Permanent permanent)
{
    var options = matches.Filter(x => x.MyDigimon == ownerSelectDigimon).ToList();
    return CardEffectCommons.IsPermanentExistsOnOpponentBattleAreaDigimon(permanent, card)
        && options.Exists(x => x.OpponentDigimon == permanent);
}

#endregion

if (ValidDigimon() && CardEffectCommons.HasMatchConditionPermanent(CanSelectOwnerDigimon))
{
    #region Select Digimon to Compare

    SelectPermanentEffect selectPermanentEffect = GManager.instance.GetComponent<SelectPermanentEffect>();

    selectPermanentEffect.SetUp(
        selectPlayer: card.Owner,
        canTargetCondition: CanSelectOwnerDigimon,
        canTargetCondition_ByPreSelecetedList: null,
        canEndSelectCondition: null,
        maxCount: 1,
        canNoSelect: false,
        canEndNotMax: false,
        selectPermanentCoroutine: SelectPermanentCoroutine,
        afterSelectPermanentCoroutine: null,
        mode: SelectPermanentEffect.Mode.Custom,
        cardEffect: activateClass);

    selectPermanentEffect.SetUpCustomMessage("Select 1 of your Digimon.", "The opponent is selecting 1 Digimon");

    IEnumerator SelectPermanentCoroutine(Permanent permanent)
    {
        ownerSelectDigimon = permanent;
        yield return null;
    }

    yield return ContinuousController.instance.StartCoroutine(selectPermanentEffect.Activate());

    #endregion

    #region Select Digimon To Delete

    if (ownerSelectDigimon != null)
    {
        SelectPermanentEffect selectPermanentEffect1 = GManager.instance.GetComponent<SelectPermanentEffect>();

        selectPermanentEffect.SetUp(
            selectPlayer: card.Owner,
            canTargetCondition: CanSelectOpponentDigimon,
            canTargetCondition_ByPreSelecetedList: null,
            canEndSelectCondition: null,
            maxCount: 1,
            canNoSelect: false,
            canEndNotMax: false,
            selectPermanentCoroutine: null,
            afterSelectPermanentCoroutine: null,
            mode: SelectPermanentEffect.Mode.Destroy,
            cardEffect: activateClass);

        selectPermanentEffect1.SetUpCustomMessage("Select 1 Digimon to delete.", "The opponent is selecting 1 Digimon to delete");

        yield return ContinuousController.instance.StartCoroutine(selectPermanentEffect.Activate());
    }

    #endregion
}
`

CardEffect/ST22/Red/ST22_08.cs
BennAbbot: I could be wrong but it looks like `card.CanLinkToTargetPermanent` compares to the card's internal linkCondition which I assume matchs the one we added, hard to tell without being able to run and test.

CardEffect/ST22/Red/ST22_08.cs
BennAbbot: bit confused which bool function you mean? I was only using this to check if we should do the effect in the first place, I wasn't using this in the selection.

Michael8818 (approved):
going to approve this for testing, as nothing looks really off to me. However this code base is always really finicky, wouldn't surprise me if something popped up from this change to the structure.

PR #652, opened 2025-12-06
feat: #644: Implement P-223

### LINKED ISSUE #644: P-223 - BT24 Box Topper
Image

hooperk:
Addresses #644

### DIFF
--- a/CardEffect/P/Purple/P_223.cs
+++ b/CardEffect/P/Purple/P_223.cs
@@ -1,0 +1,485 @@
+using System;
+using System.Collections;
+using System.Collections.Generic;
+
+namespace DCGO.CardEffects.P
+{
+    // Kuzuhamon
+    public class P_223 : CEntity_Effect
+    {
+        public override List<ICardEffect> CardEffects(EffectTiming timing, CardSource card)
+        {
+            List<ICardEffect> cardEffects = new List<ICardEffect>();
+
+            #region Alt Digivolution Condition
+
+            if (timing == EffectTiming.None)
+            {
+                bool PermanentCondition(Permanent targetPermanent)
+                {
+                    return targetPermanent.TopCard.ContainsCardName("Maid Mode") && targetPermanent.TopCard.HasLevel && targetPermanent.TopCard.Level == 6;
+                }
+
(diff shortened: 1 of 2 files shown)


Scripts/CardSource.cs
```diff
@@ -1895,6 +1895,25 @@ public bool HasLightFangOrNightClawTraits
 
     #endregion
 
+    #region [Onmyōjutsu] or [Plug-In] trait
+
+    public bool HasOnmyoOrPluginTraits
+    {
+        get
+        {
+            if (ContainsTraits("Plug-In") || ContainsTraits("PlugIn") || ContainsTraits("Plugin"))
```
hooperk: I'm not sure if I needed to be this defensive, but just in case and it can be trimmed down if unneeded

CardEffect/P/Purple/P_223.cs
```diff
@@ -0,0 +1,465 @@
...
+                cardEffects.Add(activateClass);
+
+                string EffectDiscription()
+                {
+                    return "[On Play] [When Digivolving] You may use 1 [Onmyōjutsu] or [Plug-In] trait Option card from your hand or trash without paying the cost.";
+                }
+
+                bool CanUseCondition(Hashtable hashtable)
+                {
+                    return CardEffectCommons.CanTriggerOnPlay(hashtable, card) || CardEffecCommons.CanTriggerWhenDigivolving(hashtable, card);
```
hooperk: On Play and When Digivolving can be easily combined as they have the same timing. Thought it would reduce code duplication to do so.

Scripts/CardSource.cs
x89rt2: You do not. No card exists until now with those traits. Just have them as written on the card.

CardEffect/P/Purple/P_223.cs
x89rt2: This doesn't work. You can share just about everything else but cantrigger need to be unique to the specific condition that trigger them.

x89rt2:
Both SetUpActivateClass need to be set up as true instead of false since both effects are "may". Although keep in mind you need to separate the OP from the WD.

Michael8818 (approved):

PR #657, opened 2025-12-07
Implement BT24-100

### LINKED ISSUE #592: BT24-100
Image

hooperk:
Addresses #592 

Adapted from BT23-099 Kuremi Detective Agency which is almost exactly the same card but for CS trait instead of TS trait

### DIFF
--- a/CardEffect/BT24/White/BT24_100.cs
+++ b/CardEffect/BT24/White/BT24_100.cs
@@ -1,0 +1,101 @@
+using System.Collections;
+using System.Collections.Generic;
+
+// In-Between Theater
+namespace DCGO.CardEffects.BT24
+{
+    public class BT24_100 : CEntity_Effect
+    {
+        public override List<ICardEffect> CardEffects(EffectTiming timing, CardSource card)
+        {
+            List<ICardEffect> cardEffects = new List<ICardEffect>();
+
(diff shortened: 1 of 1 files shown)


CardEffect/BT24/White/BT24_100.cs
```diff
@@ -0,0 +1,101 @@
...
+                }
+
+                bool CanUseCondition(Hashtable hashtable)
+                {
+                    return CardEffectCommons.CanTriggerOptionMainEffect(hashtable, card);
+                }
+
+                bool CanSelectCardCondition(CardSource cardSource)
+                {
+                    return cardSource.HasTSTraits;
```
x89rt2, line 53: Does this really work? My first impression is no but I would love a confirmation.

hooperk: CardSource has `HasTSTraits` defined as EqualsTraits("TS"), I'm not sure why this wouldn't work? It's based off Kuremi Detective Agency but checking for TS instead of CS. Should work the same

x89rt2: Ok yea, looks fine. Been looking at a lot of older cards lately and they did it differently so it threw me off.

Michael8818 (approved):

PR #668, opened 2025-12-09
Fix ST22-08 Link Can Target Token

BennAbbot:


CardEffect/ST22/Red/ST22_08.cs
```diff
@@ -250,7 +250,7 @@ AddLinkConditionClass LinkCondition(CardSource card)
             static bool PermanentCondition(Permanent targetPermanent)
             {
                 CardSource Top = targetPermanent.TopCard;
-                return Top.IsDigimon && Top.Level >= 3;
+                return Top.IsDigimon && Top.HasLevel && Top.Level >= 3;
```
hooperk, line 253: Can add !isToken as well? Otherwise Diaboromon Token is a valid target

BennAbbot: True, this will need to be updated for the other plug-ins as well

PR #670, opened 2025-12-09
Fix for ST22-09 Main/Sec Effects Not Triggering

BennAbbot:
I assume im somewhat at fault for this because I changed the file layout in ST22-08 and someone copied and pasted some of that code


Michael8818 (approved):

x89rt2:
I see there were issues with ST22-10, is this one fine or does it need further changes?

PR #671, opened 2025-12-09
Fix ST22-10 Main/Sec/Sec Faceup Not Triggering

BennAbbot:
I assume Im somewhat at fault for this because I changed the file layout in ST22-08 and someone copied and pasted some of that code


hooperk:
This isn't the full fix. You'll notice the wrong main was implemented, it is attempting to link, which fails for lack of link condition, instead of attempting to draw 1 and place in security. It appears to have the main effect of Sonic Shot instead

hooperk:
> Nice catch, I'll update the title since this fix only changes the effects not triggering.
> 
> Or I can remove this PR if someone is looking at a full fix.

I'll do the full fix in a few hours if nobody beats me to it. Just at work atm

hooperk:
@BennAbbot I've pushed my fix [here](https://github.com/DCGO2/DCGO-Card-Scripts/pull/674), it includes your fix to the security effects

PR #683, opened 2025-12-10
Update ST22_12.cs

x89rt2:
Attempt to fix WA gaining mem


x89rt2:
I have doubts about the - infront of mathf

hooperk:
> I have doubts about the - infront of mathf

I believe it is correct. The mathf.max is finding the cost of the linking, and ensuring the lowest it will go is 0. The minus is then taking that cost, which is positive or 0, and making it a subtraction of memory.

Michael8818 (approved):

PR #684, opened 2025-12-10
Update ST22_07.cs

x89rt2:
Line 196 getting the proper coding should maybe fix the issue of not comparing lvl to use cost


x89rt2:
https://github.com/DCGO2/DCGO-Card-Scripts/pull/686 is ahead of me

PR #686, opened 2025-12-10
ST22-07: Fix Plug-Ins not ignoring color requirement and not accounting for level

hooperk:


Scripts/Permanent.cs
```diff
@@ -1488,6 +1488,11 @@ public List<ICardEffect> EffectList_ForCard(EffectTiming timing, CardSource _car
                         {
                             if (cardEffect != null)
                             {
+                                if (cardEffect is IIgnoreColorConditionEffect)
```
hooperk, line 1493: Was not being considered because was not the topCard and was not an inherited effect.

Scripts/Permanent.cs
```diff
@@ -1488,6 +1488,24 @@ public List<ICardEffect> EffectList_ForCard(EffectTiming timing, CardSource _car
...
                             {
+                                #region Always active on card effects
+
+                                if (cardEffect is IIgnoreColorConditionEffect)
+                                {
+                                    _EffectList.Add(cardEffect);
+                                    continue;
+                                }
+
+                                if (cardEffect is IChangeCardNamesEffect)
```
hooperk, line 1499: "Rule: Name is..." Should be active for cards in all locations.
This is the only other Card Effect Type that made sense to me to always apply. 
May fix some issues with X Antibody Protoform under Phoenixmon X.

Michael8818 (approved):

PR #696, opened 2025-12-14
Implement BT24-081

### LINKED ISSUE #573: BT24-081
Image

hooperk:


CardEffect/BT24/Purple/BT24_081.cs
```diff
@@ -0,0 +1,328 @@
...
+                                }
+                            }
+
+                            return false;
+                        }
+
+                        List<AssemblyConditionElement> elements = new List<AssemblyConditionElement>(){element, element1};
+
+                        AssemblyCondition assemblyCondition = new AssemblyCondition(
+                            element: elements,
```
hooperk: Assembly does not currently support multiple discrete conditions as far as I could see, the engine will need to be updated to allow for this.
I have written this similar to how digixros is written as I expect it to end up similar, but currently this will break as it passes the wrong type.

Michael8818 (approved):

PR #710, opened 2025-12-16
Option card always gets Ignore Color Effects, even when under a permanent

hooperk:


Scripts/CardSource.cs
```diff
@@ -255,15 +255,12 @@ public bool MatchColorRequirement
 
                 #region the effects of itself
 
-                if (PermanentOfThisCard() == null)
```
hooperk, line 258: This line is meaning it does not check for the effect if under a permanent. AS this segment only checks for IgnoreColorConditions, it is safe to always apply them.

Scripts/Permanent.cs
```diff
@@ -1488,22 +1488,6 @@ public List<ICardEffect> EffectList_ForCard(EffectTiming timing, CardSource _car
...
-                                }
-
-                                if (cardEffect is IChangeCardNamesEffect)
-                                {
-                                    _EffectList.Add(cardEffect);
-                                    continue;
-                                }
-
-                                #endregion
-
```
hooperk, line 1506: Removing these, were erroneously applying IgnoreColorCondition and ChangeName to entire Permanent, not the singular card

x89rt2:
Just to verify so I can know for testing, which cards that use options from under tamers are verified to support this method?

hooperk:
> Just to verify so I can know for testing, which cards that use options from under tamers are verified to support this method?

Everything that uses option uses this method. It is the general check for color conditions

Scripts/Permanent.cs
Michael8818: keep this in mind because you are removing a core part and seems like you have only changed ignore color elsewhere you may completely break ChangeName here. It could cause some pretty major issues.

Michael8818 (approved):

PR #714, opened 2025-12-16
Update ST22_14.cs

x89rt2:
Attempting to fix the following:

"Card can't be played at 1 mem when you can meet the When would be played effect."

Not really sure if I've fixed it, copied some code from EX8 Medieval.

Have doubts on if line 59 works as is. Are "||" and/or?


Michael8818 (commented):
This still won't do it it is missing an important piece. Cards like this will require 2 timings for checking it is playable. The one that is there now is actually the effect works fine. The one missing is for the "predictive memory" basically if a card is playable the card highlights in hand to show you can use it. This will only work if it also accounts for reduced play code. check out some other cards besides medieval to find one that has 2 cost reduction effects in use.

hooperk:
> Have doubts on if line 59 works as is. Are "||" and/or?

"||" is inclusive or, yes

Michael8818 (approved):

PR #723, opened 2025-12-16
Implement P-214

### LINKED ISSUE #608: P-214 - OST Vol1 2026
Image

hooperk:


x89rt2:
Do you not need all the stuff from top?

using System.Collections;
using System.Collections.Generic;
using UnityEngine;
using System.Linq;
using Photon;
using System;
using Photon.Pun;

hooperk:
> Do you not need all the stuff from top?
> 
> using System.Collections; using System.Collections.Generic; using UnityEngine; using System.Linq; using Photon; using System; using Photon.Pun;

According to my IDE, no, not in individual cards. You can actually see them absent on some other cards, like some Promos

Michael8818:
Those are called usings, they are used to "import" other scripting references into your script as needed. it depends on what is used in the file. 

For example Math.min uses the System namespace so you need to have using System; at the top of the file. using UnityEngine is a pretty standard must have though.

Michael8818 (changes requested):
add these usings back to top:

using System.Collections;
using System.Collections.Generic;
using UnityEngine;
using System;

hooperk:
> add these usings back to top:
> 
> using System.Collections; using System.Collections.Generic; using UnityEngine; using System;

The CardEffects should never need Unity as they don't directly interact with the Unity engine, only a class calling functions of the engine would need the import. 
I also suspect some of these are being imported implicitly by the project file `<ImplicitUsings>enable</ImplicitUsings>` which removes the need to declare System, System.Colletions and System.Collections.Generic at a minimum, but I cannot find the project file to verify

Michael8818 (approved):

PR #729, opened 2025-12-18
Create P_217.cs

### LINKED ISSUE #611: P-217 - OST Vol1 2026
Image

x89rt2:
Only missing the "linked to a trait" part and that added to the can use

### DIFF
--- a/CardEffect/P_217.cs
+++ b/CardEffect/P_217.cs
@@ -1,0 +1,132 @@
+using System;
+using System.Collections;
+using System.Collections.Generic;
+using System.Linq;
+using UnityEngine.XR;
+
+//Haru Shinkai
+namespace DCGO.CardEffects.P
+{
+    public class P-217 : CEntity_Effect
+    {
+        public override List<ICardEffect> CardEffects(EffectTiming timing, CardSource card)
+        {
+            List<ICardEffect> cardEffects = new List<ICardEffect>();
+
+            #region On Play
+
+            if (timing == EffectTiming.OnPlay)
+            {
+                ActivateClass activateClass = new ActivateClass();
+                activateClass.SetUpICardEffect("Reveal 3, add 2, bot deck rest", null, card);
+                activateClass.SetUpActivateClass(null, ActivateCoroutine, -1, false, EffectDiscription());
+                cardEffects.Add(activateClass);
+                
+                string EffectDiscription()
+                {
+                    return "[On Play] Reveal the top 3 cards of your deck, Add 1 [Social] trait card and 1 [Creation], [Navi] or [Tool] trait card among them to the hand. Return the rest to the bottom of the deck.";
+                }
+
+                bool CanSelectCardCondition(CardSource cardSource)
+                {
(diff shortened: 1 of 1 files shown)


CardEffect/P_217.cs
```diff
@@ -0,0 +1,133 @@
...
+                {
+                    return CardEffectCommons.IsExistOnBattleArea(card)
+                        && CardEffectCommons.IsOwnerTurn(card))
+                        && ;
+                }
+
+                bool CanActivateCondition(Hashtable hashtable)
+                {
+                    return CardEffectCommons.IsExistOnBattleArea(card)
+                        && CardEffectCommons.CanActivateSuspendCostEffect(card));
```
hooperk: && CardEffectCommons.CanTriggerWhenLinked(hashtable, LinkPermanentCondition, null);

x89rt2: cool, I'll look into it in a sec, thanks

CardEffect/P_217.cs
```diff
@@ -0,0 +1,133 @@
...
+                }
+
+                bool LinkPermanentCondition(Permanent permanent)
+                {
+                    return ;
```
hooperk: CardEffectCommons.HasMatchConditionOwnersPermanent(permanent) && 
(permanent.TopCard.EqualsTrait("Social") || permanent.TopCard.EqualsTrait("Navi") || permanent.TopCard.EqualsTrait("Tool"));

CardEffect/P_217.cs
hooperk: CanTriggerWhenLinked: second parameter is a condition of the permanent being linked to, third parameter is a condition of the card being placed under as a link card. Since we only care about the link target, only need LinkPermanentCondition

hooperk: Also, I put this in the wrong place, it should be on CanUseCondition since it is part of the trigger condition

x89rt2: > CanTriggerWhenLinked: second parameter is a condition of the permanent being linked to, third parameter is a condition of the card being placed under as a link card. Since we only care about the link target, only need LinkPermanentCondition

So, with the understanding that the card in battle area that is getting linked to the trait doesn't matter and that the card in hand for example that I activate the link effect of needs to have the traits, is it not the second that needs null and the third that needs LinkPermanentCondition?

CardEffect/P_217.cs
x89rt2: see above question for this one as well

CardEffect/P_217.cs
hooperk: "Linked to" refers to the digimon that is getting a link card placed under it, so these effects need to check the permanent they are being attached to... but there is some weirdness on this so I'm asking in the bandai discord

hooperk: Never mind, I did have to check some weird wording but you are correct. We care that the card placed under has those traits, so it is a LinkCardCondition as the third variable, and we don't care about the second

hooperk (approved):

PR #732, opened 2025-12-18
Create P_216.cs

### LINKED ISSUE #610: P-216 - OST Vol1 2026
Image

x89rt2:


P_216.cs
```diff
@@ -0,0 +1,252 @@
...
+            #region Blocker
+
+            if (timing == EffectTiming.None)
+            {
+                cardEffects.Add(CardEffectFactory.BlockerSelfStaticEffect(isInheritedEffect: false, card: card, condition: null));
+            }
+
+            #endregion
+
+            #region On Play - Place a card under this Digimon to gain effects.
```
hooperk: Minor: Wrong description

P_216.cs
```diff
@@ -0,0 +1,252 @@
...
+
+                        bool CanUseCondition2(Hashtable hashtable)
+                        {
+                            return CardEffectCommons.IsOwnerTurn(card);
+                        }
+
+                        bool CanActivateCondition1(Hashtable hashtable)
+                        {
+                            return selectedPermanent.TopCard != null
+                                && selectedPermanent.TopCard.IsDigimon
```
hooperk: Minor since it shouldn't matter, but whatever entity type this digimon is at turn end it should still be deleted, so we  don't need to check if it is still a Digimon. Same for on delete

P_216.cs
```diff
@@ -0,0 +1,252 @@
...
+                        }
+
+                        IEnumerator ActivateCoroutine1(Hashtable _hashtable1)
+                        {
+                            yield return ContinuousController.instance.StartCoroutine(new DestroyPermanentsClass(new List<Permanent>() { selectedPermanent }, CardEffectCommons.CardEffectHashtable(activateClass1)).Destroy());
+                        }
+                        #endregion
+                    }
+                }
+            }
```
hooperk, line 149: OnPlay needs to add Can Not Digivolve

hooperk (changes requested):

hooperk (approved):

PR #737, opened 2025-12-18
Implement BT24-102

### LINKED ISSUE #594: BT24-102 Homeros

hooperk:


x89rt2:
Does this handle used up OPTs?

hooperk:
Resolves #594

CardEffect/BT24/White/BT24_102.cs
```diff
@@ -0,0 +1,237 @@
...
+                            }
+                            if (selectedEffect != null)
+                            {
+                                if (selectedEffect.EffectSourceCard != null)
+                                {
+                                    if (selectedEffect.EffectSourceCard.PermanentOfThisCard() != null)
+                                    {
+                                        Hashtable effectHashtable = CardEffectCommons.WhenDigivolvingCheckHashtableOfCard(selectedEffect.EffectSourceCard);
+
+                                        if (selectedEffect.CanUse(effectHashtable))
```
hooperk: @x89rt2 Part of CanUse() checks Once Per Turn

x89rt2: Does it handle 2 OP with OPTs and selecting the one that doesn't have it used up? This stuff I'm not familiar with yet so would love to know in case I ever have to copy from this code.

hooperk: So it isn't actually filtering them out of the choices list, it just prevents it from activating after being chosen. So you would have to know which one had or hadn't used the OPT. In theory it is valid to choose one that will fail to activate, but given this is optional anyway it'd be better if we filter that for the player. I'll see if I can work that one out tomorrow

hooperk: OK, I've put the same check in the filter creating the list of options, and also realised I forgot to ensure it worked properly for On Play as well as When Digivolving. Should now do either correctly

CardEffect/BT24/White/BT24_102.cs
```diff
@@ -0,0 +1,251 @@
...
+        public override List<ICardEffect> CardEffects(EffectTiming timing, CardSource card)
+        {
+            List<ICardEffect> cardEffects = new List<ICardEffect>();
+
+            #region Start of your Main
+            if (timing == EffectTiming.OnStartMainPhase)
+            {
+                ActivateClass activateClass = new ActivateClass();
+                activateClass.SetUpICardEffect("Gain 1 Memory. If 5+ Memory, suspend and Draw 1.", CanUseCondition, card);
+                activateClass.SetUpActivateClass(CanActivateCondition, ActivateCoroutine, -1, true, EffectDiscription());
```
Michael8818: isOptional is set to true, this effect isn't optional

CardEffect/BT24/White/BT24_102.cs
```diff
@@ -0,0 +1,251 @@
...
+                }
+
+                bool Condition()
+                {
+                    return CardEffectCommons.IsExistOnBattleArea(card);
+                }
+
+                bool PermanentCondition(Permanent permanent)
+                {
+                    return CardEffectCommons.IsPermanentExistsOnOwnerBattleArea(permanent, card) &&
```
Michael8818: use IsPermanentExistsOnOwnerBattleAreaDigimon

CardEffect/BT24/White/BT24_102.cs
```diff
@@ -0,0 +1,251 @@
...
+                string EffectDiscription()
+                {
+                    return "[End of Your Turn] By suspending this Tamer, you may activate 1 [On Play] or [When Digivolving] effect of 1 of your [Olympos XII] trait Digimon.";
+                }
+
+                bool CanSelectPermanentCondition(Permanent permanent)
+                {
+                    return CardEffectCommons.IsPermanentExistsOnOwnerBattleAreaDigimon(permanent) &&
+                        permanent.TopCard.EqualsTraits("Olympos XII") &&
+                        selectedPermanent.EffectList(EffectTiming.OnEnterFieldAnyone).Any(CanBeEffectCandidate);
```
Michael8818, line 92: This just checks if the card uses the on enterfield timing, so would be true for any All turn effects that use the on enter field timing. Need to confirm it is also an On Play or When Digivolving effect specifically.

Michael8818 (changes requested):

CardEffect/BT24/White/BT24_102.cs
hooperk: CanBeEffectCandidate then checks for IsWhenDigivolving or IsOnPlay within the Any(), this just narrows down which list of effects it checks

Michael8818 (approved):

PR #738, opened 2025-12-18
Factory methods for common ActivateClasses and Shared Effects

hooperk:
I find when I create cards that share an effect, I am generally using the exact same boiler plate every time for the activate class and CanUseCondition declarations, so I've extracted them to a factory method which will add them to the passed cardEffects. I've made the changes to several recently added cards to show how it would work

### DIFF
--- a/CardEffect/BT24/Yellow/BT24_036.cs
+++ b/CardEffect/BT24/Yellow/BT24_036.cs
@@ -92,49 +92,15 @@ namespace DCGO.CardEffects.BT24
                     }
                 }
 
-            #endregion
-
-            #region On Play
-            if (timing == EffectTiming.OnEnterFieldAnyone)
-            {
-                ActivateClass activateClass = new ActivateClass();
-                activateClass.SetUpICardEffect(SharedEffectName(), CanUseCondition, card);
-                activateClass.SetUpActivateClass(CanActivateCondition, (hashTable) => SharedActivateCoroutine(hashTable, activateClass), -1, false, SharedEffectDescription("On Play"));
-                cardEffects.Add(activateClass);
-
-                bool CanUseCondition(Hashtable hashtable)
-                {
-                    return CardEffectCommons.IsExistOnBattleAreaDigimon(card) &&
-                        CardEffectCommons.CanTriggerOnPlay(hashtable, card);
-                }
-
-                bool CanActivateCondition(Hashtable hashtable)
-                {
-                    return CardEffectCommons.IsExistOnBattleAreaDigimon(card) &&
-                        CardEffectCommons.HasMatchConditionPermanent(SharedCanSelectPermanentCondition);
-                }
-            }
-            #endregion
-
-            #region On Deletion
-
-            if (timing == EffectTiming.OnDestroyedAnyone)
-            {
-                ActivateClass activateClass = new ActivateClass();
-                activateClass.SetUpICardEffect(SharedEffectName(), CanUseCondition, card);
-                activateClass.SetUpActivateClass(CanActivateCondition, (hashTable) => SharedActivateCoroutine(hashTable, activateClass), -1, false, SharedEffectDescription("On Deletion"));
-                cardEffects.Add(activateClass);
-
-                bool CanUseCondition(Hashtable hashtable)
-                {
-                    return CardEffectCommons.CanTriggerOnDeletion(hashtable, card);
-                }
-
-                bool CanActivateCondition(Hashtable hashtable)
-                {
-                    return CardEffectCommons.CanActivateOnDeletion(card);
-                }
-            }
+            CardEffectFactory.ActivateClassesForSharedEffects(cardEffects, timing, card,
+                                                              effectName: SharedEffectName(),
+                                                              activateCoroutine: SharedActivateCoroutine,
+                                                              effectDescription: SharedEffectDescription,
+                                                              optional: false,
+                                                              maxCountPerTurn: -1,
+                                                              onPlay: true,
+                                                              onDeletion: true);
+                                                              
             #endregion
 
             #region Link
--- a/Scripts/CardEffectFactory.cs
+++ b/Scripts/CardEffectFactory.cs
@@ -627,4 +627,581 @@ public partial class CardEffectFactory
     }
 
     #endregion
+
+    #region Shared Effect Condition Creators
+
+    /// Method that can be used to quickly set up effects with multiple trigger conditions
+    /// Example: [When Digivolving] [When Attacking] (Do Something) can be set up with whenDigivolving = true and whenAttacking = true
+    /// Does not support sharing with Link or Inherited effect, purposefully to ensure they don't accidentally share a hashValue on Once Per Turn effects
+    /// 
+    /// Always pass in the cardEffects, timing and card variables from the calling card script, this ensures the below methods are added to that card's effects for the correct timing
+    /// 
+    /// The minimal conditions for triggering and activating the timings are included automatically,
+    /// further checks can be added to all created CanUseConditions and CanActivateConditions by passing those additonal checks with additionalUseCondition and additionalActivateCondition
+    /// For example, if a effect read "[On Play] [When Attacking] By trashing 1 card in hand..." you might create 
+    /// bool AdditionalActivateCondition(Hashtable hashtable) => card.Owner.HandCards.Count >= 1;
+    /// and pass such a method into this function.
+    /// 
+    /// By default, the effects are no optional and have no max per turn count. You can omit those parameters when this is the case and only change them when needed.
+    /// To ensure correct behaviour of X Per Turn, the hashValue is shared as the effect is once per turn regardless of how it is triggered.
+    /// 
+    /// Finally, for whichever combination of effects you want, you will set the corresponding booleans to true. Order does not matter when refering to them by name
+    /// Examples: 
+    /// "[when Moving] [When Digivolving] [When Attacking] ..." => (... whenMoving: true, whenDigivolving: true, whenAttacking: true)
+    /// "[When Digivolving] [On Deletion] ..." => (... whenDigivolving: true, onDeletion: true)
+    /// "[End of Attack] [End of Opponent's Turn] ..." => (... endOfAttack: true, endOfOpponentTurn: true)
+    public static List<ICardEffect> ActivateClassesForSharedEffects(ref List<ICardEffect> cardEffects,
+                                                                    EffectTiming timing, 
+                                                                    CardSource card,
+                                                                    string effectName, 
+                                                                    Func<Hashtable, ActivateClass, IEnumerator> activateCoroutine, 
+                                                                    Func<string, string> effectDescription,
+                                                                    bool optional,
+                                                                    Func<Hashtable, bool> additionalUseCondition = null,
+                                                                    Func<Hashtable, bool> additionalActivateCondition = null, 
+                                                                    int maxCountPerTurn = -1,
+                                                                    string hashValue = null,
+                                                                    bool whenMoving = false,
+                                                                    bool onPlay = false,
+                                                                    bool whenDigivolving = false,
+                                                                    bool whenAttacking = false,
+                                                                    bool onDeletion = false,
+                                                                    bool whenLinking = false,
+                                                                    bool endOfAttack = false,
+                                                                    bool endOfYourTurn = false,
+                                                                    bool endOfOpponentTurn = false,
+                                                                    bool endOfAllTurns = false,
+                                                                    bool startOfYourMainPhase = false)
+    {
+        if (whenMoving && timing == EffectTiming.OnMove)
+        {
+            cardEffects.Add(WhenMovingClass(card, effectName, activateCoroutine, effectDescription("When Moving"), optional, additionalUseCondition, additionalActivateCondition, maxCountPerTurn, hashValue));
+        }
+
+        if (onPlay && timing == EffectTiming.OnEnterFieldAnyone)
+        {
+            cardEffects.Add(OnPlayClass(card, effectName, activateCoroutine, effectDescription("On Play"), optional, additionalUseCondition, additionalActivateCondition, maxCountPerTurn, hashValue));
+        }
+
+        if (whenDigivolving && timing == EffectTiming.OnEnterFieldAnyone)
+        {
+            cardEffects.Add(WhenDigivolvingClass(card, effectName, activateCoroutine, effectDescription("When Digivolving"), optional, additionalUseCondition, additionalActivateCondition, maxCountPerTurn, hashValue));
+        }
+
+        if (whenAttacking && timing == EffectTiming.OnAllyAttack)
+        {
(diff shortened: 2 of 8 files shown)


hooperk:
> * Rather than passing in the CardEffect List instead just return the `ActivateClass` that was created and add it to the list in `CEntity_Effect`, this makes the code a little bit more reusable/less coupled and matches the other `CardEffectFactory` functions.
Something like this is what I started with, but I quickly found what I wanted was an effect I could say I want OnPlay and WhenAttacking this time, and just have both created, or in another place declare some other combination. Especially since the CanUseCondition is always the same on the boilerplates I'm looking to not need to copy and paste. What I do think I'll do best is create the method you describe for creating one ActivateCondition, and have this one call that, best of both options, 
> * I find the name a bit confusing I think something like `CreateCommonActivateClass()` describes what it is doing a bit more, that is a bit verbose so you could just go with `CommonActivateClass()` or `ActivateClass()` since `CardEffectFactory` kind of implies that you are creating something.
I'll try think of another name, but I do want the name for this one to confer that it creates multiple effects
> * I would rearrange the parameter's so that the `int maxCountPerTurn` and `string hashValue` are the last parameters, as if I understand correctly you only need to set `hashValue`  if `maxCountPerTurn` is set, since to me these are the least used params I would put them last so that you dont always need to set them.
Good point

hooperk:
> The main reason I want the singular functions like `CommonActivateClass` and `WhenDigivolvingActivateClass` is because I think it would be usefully to cut down boiler plate in effects that both do and don't share effects.
> 
> But if you don't want to make that change and only want to focus on just the function that creates multiple effects, dont feel pressured I can add those myself later.

I actually do like this, and will add it to this branch later on when I have time to work on it. At a minimum that could be 4 less lines of copy paste at the top of every definition. The one thing is that the timing can't be passed. If we calculate on the timing inside a method returning ActivateClass then we risk returning null when the timing doesn't match, and I don't want to fill CardEffects with nulls.

hooperk:
> Ah I missed the part of it making multiple class, I do think that would also be handy as its own function, if that's the case I would recommend using an Enum flags instead of multiple bools. I would also recommend that this be the first param in the function and not an optional param.
> 
> I think this is how you do it in c# but to im not to familiar with it
> 
> ```cs
> [Flags]
> public enum Triggers
> {
>     None    = 0,
>     WhenDigi = 1 << 0,
>     OnPlay = 1 << 1, 
>     WhenMoving = 1 << 2,
>     WhenAttacking = 1 << 3
> }
> 
> Triggers triggers = Triggers.WhenDigi  | Triggers.WhenAttacking;
> 
> if(triggers .HasFlag(Triggers.WhenAttacking)
> {
> }
> ```

This is definitely the more sophisticated approach, though I worry it is slightly more difficult for new contributors than just setting the optional bools to true. Part of this is trying to make it as easy to use as possible

hooperk:
> Up to you I just find this way easier to extend since you don't need keep adding params to the end of the function and less verbose since if you only need 2 flags you dont need to set a bunch of bools to false, less likely to make a mistake and cause a bug to.

Yeah, it's just one of those places where the natural English statement is "I want this enabled for When Moving **and** On Play" but then you need to do "|" instead of "&". Similar to bugs that show up sometimes because we need to use Math.min when setting a maxCount.

Since they are optional parameters I don't think it is too much effort to add them, since they won't break any existing usages. also why I'm keeping those at the very end.

Scripts/CardEffectFactory.cs
```diff
@@ -554,4 +554,224 @@ IEnumerator ActivateCoroutine(Hashtable hashtable)
...
+        }
+
+        bool CanActivateCondition(Hashtable hashtable)
+        {
+            return CardEffectCommons.CanActivateOnDeletion(card) && 
+                (additionalActivateCondition == null || additionalActivateCondition);
+        }
+    }
+        
+    #endregion
```
hooperk, line 914: There could be many more, but these 5 fit my original use case of shared trigger conditions and are the most common triggers. I don't want to make a SelfTrashedCard class if it doesn't come up very often, for example. But It can be extended by anyone and I would welcome someone else making a SelfTrashedCardClass, or others.

x89rt2:
Twice per turns are currently OPT so I'm down to a new method to keep track of those.

hooperk:
That could even be SetMaxCountPerTurn(int count, string hashValue) to ensure its never forgotten to add the hashValue.

Also are you pushing back or making sure I remember Linked as well?
My defense of these is that having them all in this method represents almost a single checklist of "have I addressed all of the variables this expects?"

x89rt2:
There is one other issue that needs to be discuss if we are to implement this. We currently have an issue where copy effects like the gamma or bagra archetypes, if they have 2 of the same effect in sources and they us hashstrings, once one of them gets used up, the other does so as well. It would be nice if this new method supported whatever fix that bug requires.

Or we can kick that can down the road.

hooperk:
Hmm. That will be a deeper issue since the effects are copied to actually be on the top card, which means then both effects are on the same cardSource with the same hashString, which is exactly when we would usually mark them as the same effect.
Fixing that would involve a larger change of somehow counting the source card as the source of the effect rather than just copying it to the top card.

x89rt2:
kick the can it is

Scripts/CardEffectFactory.cs
```diff
@@ -559,4 +559,390 @@ IEnumerator ActivateCoroutine(Hashtable hashtable)
...
+
+    #endregion
+
+    #region Start/End of Turn/Phase, Your/Opponent's/All Turns
+
+/**
+ *  Can be used for Your Turn, Opponent's Turn, All Turns, End of [Your|Opponent's|All] Turn, Start of Turn/Phase effects 
+ *  based on the timing it is used at and additional use conditions passed
+ */
+    public static ActivateClass TurnTimingClass(CardSource card,
```
hooperk, line 1012: Started writing this as "End of Your Turn", then it became end of either turn as determined by yourTurn and Opponent's Turn parameters... and then I realized that the only difference between a start of your turn, end of your turn and start of main effect is the timing it is added for. Likewise a YourTurn, Opponent's Turn and All Turns When something happens effect is also the same, with the additional triggering condition. Does mean it is hard to name this one. Might add additional methods for those variations in another commit

Scripts/CardEffectFactory.cs
```diff
@@ -583,34 +583,46 @@ public static List<ICardEffect> ActivateClassesForSharedEffects(ref List<ICardEf
                                                                     bool whenDigivolving = false,
                                                                     bool whenAttacking = false,
                                                                     bool onDeletion = false,
-                                                                    bool whenLinking = false)
+                                                                    bool whenLinking = false,
+                                                                    bool endOfAttack = false,
+                                                                    bool endOfYourTurn = false,
+                                                                    bool endOfOpponentTurn = false,
+                                                                    bool endOfAllTurns = false)
```
BennAbbot: This is kind of what I was worried about, this is becoming a bit to monolithic/verbose, if you don't want to use enum flags, you could also consider making these bool part of a class/struct or list of enums?

hooperk: Consider that the places that use this method will only ever have like 3 of these at most. They'll just reference those 3 by name: true

This method should never be used in a way where every single parameter is passed.

hooperk:
@x89rt2 I could use your opinion as someone likely to use this. Which of these would you find easier to use, or would you find them roughly the same? These two examples would be for creating the card effects for a shared On Play / When Digivolving / When Attacking.

CardEffectFactory.ActivateClassesForSharedEffects(cardEffects, timing, card, 
                                                                "Trash a card to delete all opponent's lowest digimon.",
                                                                SharedActivateCoroutine,
                                                                SharedEffectDescription,
                                                                optional: false,
                                                                additionalActivateCondition: AdditionalActivateCondition,
                                                                onPlay: true,
                                                                whenDigivolving: true,
                                                                whenAttacking: true);
                                                                
OR

CardEffectFactory.ActivateClassesForSharedEffects(cardEffects, timing, card, 
                                                                "Trash a card to delete all opponent's lowest digimon.",
                                                                SharedActivateCoroutine,
                                                                SharedEffectDescription,
                                                                triggers: TriggerType.OnPlay | TriggerType.WhenDigivolving | TriggerType.WhenAttacking,
                                                                optional: false,
                                                                additionalActivateCondition: AdditionalActivateCondition);

x89rt2:
> @x89rt2 I could use your opinion as someone likely to use this. Which of these would you find easier to use, or would you find them roughly the same? These two examples would be for creating the card effects for a shared On Play / When Digivolving / When Attacking.
> 
> CardEffectFactory.ActivateClassesForSharedEffects(cardEffects, timing, card, "Trash a card to delete all opponent's lowest digimon.", SharedActivateCoroutine, SharedEffectDescription, optional: false, additionalActivateCondition: AdditionalActivateCondition, onPlay: true, whenDigivolving: true, whenAttacking: true);
> 
> OR
> 
> CardEffectFactory.ActivateClassesForSharedEffects(cardEffects, timing, card, "Trash a card to delete all opponent's lowest digimon.", SharedActivateCoroutine, SharedEffectDescription, triggers: TriggerType.OnPlay | TriggerType.WhenDigivolving | TriggerType.WhenAttacking, optional: false, additionalActivateCondition: AdditionalActivateCondition);

I think I need an example of a card to fully visualize this. Can you do BT24-036 in this style ignoring the alt evo and the link stuff so I can visualize it?

CardEffect/BT24/Yellow/BT24_036.cs
```diff
@@ -92,49 +92,23 @@ IEnumerator SelectPermanentCoroutine(Permanent permanent)
...
-            #endregion
+            CardEffectFactory.ActivateClassesForSharedEffects(cardEffects, timing, card,
+                                                              effectName: SharedEffectName(),
+                                                              activateCoroutine: SharedActivateCoroutine,
+                                                              effectDescription: SharedEffectDescription,
+                                                              optional: false,
+                                                              maxCountPerTurn: -1,
+                                                              onPlay: true,
+                                                              onDeletion: true);
+            //TODO: Remove one of these methods
```
hooperk: @x89rt2 Here are the two examples of the function that will add the On Play and On Deletion for this shared methods to the card. Above this line is one method, and below this line the other. I'd like your opinion on which you feel is more straightforward, particularly about how we are flagging that it is onPlay and onDeletion. Above has an optional parameter for each trigger type, below uses an enum, and a .Multiple method to list any number of them.

CardEffect/BT24/Yellow/BT24_036.cs
```diff
@@ -92,49 +92,23 @@ IEnumerator SelectPermanentCoroutine(Permanent permanent)
...
-                bool CanUseCondition(Hashtable hashtable)
-                {
-                    return CardEffectCommons.IsExistOnBattleAreaDigimon(card) &&
-                        CardEffectCommons.CanTriggerOnPlay(hashtable, card);
-                }
-
-                bool CanActivateCondition(Hashtable hashtable)
-                {
-                    return CardEffectCommons.IsExistOnBattleAreaDigimon(card) &&
-                        CardEffectCommons.HasMatchConditionPermanent(SharedCanSelectPermanentCondition);
```
hooperk, line 114: Purposefully removed this line to allow the player to whiff as well

CardEffect/BT24/Yellow/BT24_036.cs
x89rt2: Did you write up a guide for it somewhere? How to hash, denote once or twice per turn, make optional or mandatory, what if I want to add something on the canuse or canactivate such as if you have a card in hand?

x89rt2: oh, I see the opt and the optional: false, above

is it maybe additionalActivateCondition and additionalUseCondition?

also, preference between both methods is the top one. easier to copy and paste.

hooperk: I haven't yet, but I'll put it as a documentation comment above the method in CardEffectFactory. But you are correct, additionalUseCondition and additionalActivateCondition are for adding any additional checks beyond the ones that are always needed for each trigger type. So if we want to check they have at least 1 card in hand before they may activate, we would define
bool AdditionalActivateCondition(Hashtable hash) { return card.Owner.HandCards.Count >=1; }
and include additionalActivateCondition: AdditionalActivateCondition

x89rt2: maxcountperturn could probably be internal. If it's not mentioned it's -1, if it's added, then it has an x per turn.

hooperk: It is optional, I only showed it here so as not to confuse things. same as optional

Scripts/CardEffectFactory.cs
```diff
@@ -559,4 +559,575 @@ IEnumerator ActivateCoroutine(Hashtable hashtable)
...
+    /// and pass such a method into this function.
+    /// 
+    /// By default, the effects are no optional and have no max per turn count. You can omit those parameters when this is the case and only change them when needed.
+    /// To ensure correct behaviour of X Per Turn, the hashValue is shared as the effect is once per turn regardless of how it is triggered.
+    /// 
+    /// Finally, for whichever combination of effects you want, you will set the corresponding booleans to true. Order does not matter when refering to them by name
+    /// Examples: 
+    /// "[when Moving] [When Digivolving] [When Attacking] ..." => (... whenMoving: true, whenDigivolving: true, whenAttacking: true)
+    /// "[When Digivolving] [On Deletion] ..." => (... whenDigivolving: true, onDeletion: true)
+    /// "[End of Attack] [End of Opponent's Turn] ..." => (... endOfAttack: true, endOfOpponentTurn: true)
```
hooperk, line 652: @x89rt2 added the docs which hopefully explain the function well, and there's now 7 examples of cards using the method in this PR

Scripts/CardEffectFactory.cs
```diff
@@ -647,6 +648,10 @@ public static List<ICardEffect> ActivateClassesForSharedEffects(ref List<ICardEf
         {
             cardEffects.Add(TurnTimingClass(card, effectName, activateCoroutine, effectDescription("End of All Turns"), optional, additionalUseCondition, additionalActivateCondition, maxCountPerTurn, yourTurn: true, opponentTurn: true));
         }
+        if(startOfYourMainPhase && timing == EffectTiming.OnStartMainPhase)
+        {
+            cardEffects.Add(TurnTimingClass(card, effectName, activateCoroutine, effectDescription("Start of Your Main Phase"), optional, additionalUseCondition, additionalActivateCondition, maxCountPerTurn, yourTurn: true, opponentTurn: false));
+        }
```
x89rt2, line 722: can we auto add a can use requirement of your turn on this one? same for any similar ones that might be missing it

hooperk: Yes, that is what yourTurn: true does

hooperk: The TurnTiming Class uses the yourTurn and opponentTurn bools to add to the canUseRequirements check that it can only be used on the turns that have been indicated

hooperk: Made it even more explicit, this now use StartOfYourMainPhaseClass, which in turn sets up to perform a check that it is your turn

Michael8818:
I think this should be merged post new repo. So it can be tested before we actually merge

PR #739, opened 2025-12-18
Update BT16-088

hooperk:
fixed able to de-digivolve immune digimon


Michael8818:
What is this change for?

hooperk:
There was an issue reported that this was able to de-digivolve an immune digimon. I believe it is because it only checks if tamer effects are allowed effect the target, and never checks immunity to DeDigivolving

x89rt2:
Nevermind, this might not be an issue. Gallant X is only digimon immune and the other gallant is a your turn only. I'd find it very unlikely for the reported issue to come up naturally in a normal game so we can drop this one.

x89rt2:
Closing until we get a legit report.

PR #814, opened 2025-12-19
Is Max Cost - Adjustment

hooperk:
GetNonMaxCostPermanents was including cards without play cost instead of guaranteeing that they are never considered as having the max cost, as they have no cost.


hooperk:
Fixes part of the issues mentioned in [#3](https://github.com/DCGO2/DCGO-Card-Scripts/issues/3)

LewisWelch94 (approved):

PR #818, opened 2025-12-19
Updating BT20-081

hooperk:
Effect was not counting as a Digimon Effect so ignoring Shakkoumon immunity


x89rt2:
I'm not convinced by this one. There's too few cards with this code but there are SOME cards with this code. I'll leave it to someone who knows more.

Michael8818:
Not a correct fix, and testers can not confirm issue. Closing for now

PR #824, opened 2025-12-20
Fix D Reaper being a valid target for IsMinCost

hooperk:
HasCost, HasUseCost and HasPlayCost check if Cost >= 0. This is correct in case we ever get a hypothetical 0 cost. However, our default on the base entity is 0 instead of -1, meaning if it is just not defined we technically HaveCost. Tokens explicitly set it to -1, I cannot check if the same is true for Mother D Reaper, but I suspect it just is not defined.

### DIFF
--- a/Scripts/CEntity_Base.cs
+++ b/Scripts/CEntity_Base.cs
@@ -330,8 +330,8 @@ public class CEntity_Base : ScriptableObject
     }
     #endregion
 
-    #region �R�X�g������
-    public bool HasCost => cardKind != CardKind.DigiEgg && PlayCost >= 0;
+    #region Whether the card has any cost
+    public bool HasCost => PlayCost >= 0;
     #endregion
 
     #region Wheter the card has play cost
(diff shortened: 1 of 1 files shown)


hooperk:
Fixes part of the issues mentioned in [#3](https://github.com/DCGO2/DCGO-Card-Scripts/issues/3)

Scripts/CEntity_Base.cs
```diff
@@ -330,7 +330,7 @@ public bool HasLevel
     }
     #endregion
 
-    #region �R�X�g������
+    #region �R�X�g������
     public bool HasCost => cardKind != CardKind.DigiEgg && PlayCost >= 0;
```
Michael8818: remove the cardKind != CardKind.DigiEgg check but keep the Playcost >= 0

Michael8818 (changes requested):

Michael8818 (approved):

PR #825, opened 2025-12-20
Fix Deck Editor not displaying Mother Eater Cost

hooperk:


hooperk:
Local testing of this would also be able to verify if #824 is correct

Scripts/DetailCard_DeckEditor.cs
```diff
@@ -140,7 +140,7 @@ public async void SetUpDetailCard(CEntity_Base cEntity_Base)
         cardID_RarityText.text = $"{cEntity_Base.CardID} {cEntity_Base.rarity}";
 
         //プレイコスト
-        if (cEntity_Base.cardKind != CardKind.DigiEgg)
+        if (cEntity_Base.PlayCost >= 0)
```
Michael8818: check cEntity_Base.HasCost instead (checks same thing).

Also add this check to line 208 of CardPrefab_CreateDeck.cs

Michael8818 (changes requested):

Michael8818 (approved):

PR #826, opened 2025-12-20
Allow Non-Digimon entities to show any Traits they have

hooperk:


hooperk:
This one I am less certain of than the others, I believe it should make traits show for tamers, options etc, but I cannot guarantee it doesn't overlap something else because it is expecting a Digimon details object to have left some screen real estate in a particular location.

Should be possible to verify on a local machine by anyone with the front end, as it does not need to connect to a game

Michael8818:
@hooperk You were on the right track here, it was just a little more involved in the code to get the UI to display correctly. I have updated it on my end and pushed that up so I am just going to close this PR as done without merging.

PR #832, opened 2025-12-21
Create P_224.cs

### LINKED ISSUE #645: P-224 - BT24 Box Topper
Image

x89rt2:


CardEffect/P/Black/P_224.cs
```diff
@@ -0,0 +1,350 @@
...
+
+                        IEnumerator SelectPermanentCoroutine(Permanent permanent)
+                        {
+                            Permanent selectedPermanent = permanent;
+
+                            if (selectedPermanent != null)
+                            {
+                                yield return ContinuousController.instance.StartCoroutine(selectedPermanent.AddDigivolutionCardsBottom(new List<CardSource>() { selectedCards[0] }, activateClass));
+
+                                if (CardEffectCommons.GetOwnerHandCount(card) <= 7)
```
hooperk, line 163: Add a check of this card's digivolution cards to ensure the card was placed

CardEffect/P/Black/P_224.cs
```diff
@@ -0,0 +1,350 @@
...
+                string EffectDiscription()
+                {
+                    return "[Main] By suspending this Tamer, you may play 1 level 5 or higher [Xros Heart] trait Digimon card from under any of your Tamers with the play cost reduced by 1.";
+                }
+
+                bool CanSelectCardCondition(CardSource cardSource)
+                {
+                    return !cardSource.willBeRemoveSources
+                        && cardSource.IsDigimon
+                        && cardSource.HasLevel && cardSource.Level => 5
```
hooperk: ">=", "=>" has a different meaning to C#

CardEffect/P/Black/P_224.cs
```diff
@@ -0,0 +1,350 @@
...
+                {
+                    return CardEffectCommons.IsExistOnBattleArea(card)
+                        && CardEffectCommons.CanActivateSuspendCostEffect(card)
+                        && card.Owner.GetBattleAreaPermanents().Count(CanSelectPermanentCondition) >= 1;
+                }
+
+                IEnumerator ActivateCoroutine(Hashtable hashtable)
+                {
+                    yield return ContinuousController.instance.StartCoroutine(new SuspendPermanentsClass(new List<Permanent>() { card.PermanentOfThisCard() }, CardEffectCommons.CardEffectHashtable(activateClass)).Tap());
+
```
hooperk, line 257: Add a check that this permanent is now suspended

CardEffect/P/Black/P_224.cs
```diff
@@ -0,0 +1,350 @@
...
+                        yield return StartCoroutine(selectCardEffect.Activate());
+
+                        IEnumerator SelectCardCoroutine(CardSource cardSource)
+                        {
+                            selectedCards.Add(cardSource);
+
+                            yield return null;
+                        }
+
+                        int cost = selectCard.BasePlayCostFromEntity - 1;
```
hooperk: selectedCards.get(0) and check count == 1 before referencing it

hooperk (changes requested):

CardEffect/P/Black/P_224.cs
x89rt2: several tamers don't verify this such as BT23_085 so I think it should be fine just with the CardEffectCommons.CanActivateSuspendCostEffect(card)

CardEffect/P/Black/P_224.cs
x89rt2: not familiar with this method so I used a different method

CardEffect/P/Black/P_224.cs
x89rt2: Not sure it's necessary here either. I don't believe a single effect exists that can interrupt after you've selected a card but before it gets placed under.

hooperk: Not so much for an interuptive effect as some blanket effect preventing cards from being placed under tamers/any other permanent

x89rt2: What effect allows that? I'm not familiar with any. If such an effect comes to be, we'll have major problems with older cards.

hooperk: True, Don't think there are any currently

hooperk (approved):

PR #836, opened 2025-12-21
Update ST22_07.cs

x89rt2:
fix incorrect description


x89rt2:
Also, we have an issue with the op bypassing menoa floodgate. This seems like a similar issue to bt23 shakkou at bypassing the hashstring possibly due to the shared coroutine. https://discord.com/channels/839041812932329512/1452399399190790289

x89rt2:
https://github.com/DCGO2/DCGO-Card-Scripts/pull/837

PR #838, opened 2025-12-21
Update CardEffectFactory.cs and BT23_025.cs

x89rt2:
Adding every turn timing to self deletion


x89rt2:
Changed self deletion timing on MarinAnge to every turn.

hooperk:
Nice, I was thinking something like this seemed like it might have been missing but hadn't fully worked out what yet

hooperk (approved):

x89rt2:
If this gets approved, reminder to change code on Dark Masters as well

Michael8818 (approved):

PR #839, opened 2025-12-22
Fix Added Skill effects sharing OPT even when they have different actual source cards

hooperk:


Scripts/ICardEffect.cs
```diff
@@ -92,6 +93,28 @@ public void SetEffectSourcePermanent(Permanent effectSourcePermanent)
...
+    {
+        AddedSourceCard = cardSource;
+    }
+
+    #endregion
+
+    public CardSource GetOriginalEffectSource()
+    {
+        return AddedSourceCard ?? EffectSourceCard;
+    }
```
hooperk, line 116: I do not want to mess with what EffectSourceCard provides. Instead this method can be used where we specifically know we want the AddedSourceCard if it exists. So far this is only for the OPT check, although if convenient and safe I think it would be nice if the display of "Select an effect to activate" next dialogue also showed the art of the source card

Scripts/CardEffects/AddSkillClass.cs
```diff
@@ -16,6 +16,11 @@ public List<ICardEffect> GetCardEffect(CardSource card, List<ICardEffect> getCar
     {
         if (_cardSourceCondition(card))
         {
+            List<ICardEffect> effects = _getEffects(card, getCardEffect, timing);
+            foreach(ICardEffect effect in effects)
+            {
+                effect.SetAddedSourceCard(card);
+            }
```
hooperk, line 23: When an AddedSkillClass would provide its list of added effects, it first ensures all of them have a reference to the actual card that passed them.

SetEffectSourceCard 1 line below preview is setting the source of the AddedSkillEffect, but not the skill that is passed.

hooperk:
Closing this for now, going to try just adjusting that SetEffectSource is targeting the correct effect

PR #840, opened 2025-12-22
Set Correct Source Card for Added Effects

hooperk:


Scripts/CardEffects/AddSkillClass.cs
```diff
@@ -17,10 +17,12 @@ public List<ICardEffect> GetCardEffect(CardSource card, List<ICardEffect> getCar
         if (_cardSourceCondition(card))
         {
             getCardEffect = _getEffects(card, getCardEffect, timing);
+            foreach (ICardEffect effect in getCardEffect)
+            {
+                effect.SetEffectSourceCard(card);
+            }
         }
 
-        SetEffectSourceCard(card);
```
hooperk, line 22: This was changing the EffectSourceCard of only the AddedSkillClass, not the skills it was adding

Michael8818:
@x89rt2 can I get some cards that could be tested to confirm this and some that would need confirmed based on this change?

x89rt2:
Not really sure what's happening here so @hooperk would have to provide a basic explanation for me to find cards or provide them himself.

hooperk:
@x89rt2 this is a fix for the effects of say the Gammamon level 5s with 2 of the same once per turn underneath, incorrectly expending both of them instead of just one. Or Chaosdramon with Machinedramon underneath. I don't have the full list of things that have tested that before though

hooperk:
@Michael8818 BT10-011 or BT21-077 as top card and/or via inheritable. with 2 EX10-042 for example, and add a card to digivolution cards via BT21-069.
Expected result: both EX10-042 shared effects trigger

PR #847, opened 2025-12-23
Implement BT24-033

### LINKED ISSUE #525: BT24-033
Image

hooperk:
Closing #525


x89rt2 (changes requested):
We prefer a different method for the Your Turn. Use BT20_046.cs as a guide, Espimon. The advantage is that when you hover a digimon over it, the cost prediction takes the discount into consideration.

x89rt2 (approved):

PR #868, opened 2025-12-24
Update BT23_032.cs

x89rt2:
Attempting shared AT and ESS AT effect using BT23-023 Whamon as reference.

### DIFF
--- a/CardEffect/BT23/Yellow/BT23_032.cs
+++ b/CardEffect/BT23/Yellow/BT23_032.cs
@@ -273,7 +273,7 @@ namespace DCGO.CardEffects.BT23
                 return CardEffectCommons.IsExistOnBattleAreaDigimon(card);
             }
 
-            IEnumerator SharedActivateCoroutine(Hashtable hashtable, ActivateClass activateClass)
+            IEnumerator SharedActivateCoroutine(Hashtable hashtable, ActivateClass activateClass, bool IsTopCard)
             {
                 bool CanSelectCardCondition(CardSource cardSource)
                 {
@@ -283,8 +283,9 @@ namespace DCGO.CardEffects.BT23
                         && CardEffectCommons.CanPlayAsNewPermanent(cardSource, false, activateClass);
                 }
 
-                // Using this to cover both non ESS & ESS case.
-                var thisPermament = card.PermanentOfThisCard();
+                Permanent thisPermament = null;
+                if (IsTopCard) thisPermament = card.PermanentOfThisCard();
+                else thisPermament = card.PermanentOfThisCard().TopCard.PermanentOfThisCard();
 
                 if (thisPermament.StackCards.Exists(CanSelectCardCondition))
                 {
--- a/CardEffect/BT23/Yellow/BT23_032.cs
+++ b/CardEffect/BT23/Yellow/BT23_032.cs
@@ -283,9 +283,7 @@ namespace DCGO.CardEffects.BT23
                         && CardEffectCommons.CanPlayAsNewPermanent(cardSource, false, activateClass);
                 }
 
-                Permanent thisPermament = null;
-                if (IsTopCard) thisPermament = card.PermanentOfThisCard();
-                else thisPermament = card.PermanentOfThisCard().TopCard.PermanentOfThisCard();
+                var thisPermament = card.PermanentOfThisCard();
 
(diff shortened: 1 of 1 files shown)


CardEffect/BT23/Yellow/BT23_032.cs
```diff
@@ -262,89 +262,92 @@ IEnumerator SelectPermanentCoroutine(Permanent permanent)
...
-                        && CardEffectCommons.CanTriggerWhenRemoveField(hashtable, card)
-                        && !CardEffectCommons.IsByEffect(hashtable, effect => effect.EffectSourceCard.Owner == card.Owner);
-                }
+            bool SharedCanActivateCondition(Hashtable hashtable)
+            {
+                return CardEffectCommons.IsExistOnBattleAreaDigimon(card);
+            }
 
-                bool CanActivateCondition(Hashtable hashtable)
+            IEnumerator SharedActivateCoroutine(Hashtable hashtable, ActivateClass activateClass)
```
hooperk: You'll need to change the signature to expect a boolean, even if you don't use it, for the calling methods to not error

CardEffect/BT23/Yellow/BT23_032.cs
```diff
@@ -262,89 +262,92 @@ IEnumerator SelectPermanentCoroutine(Permanent permanent)
...
-                    return CardEffectCommons.IsExistOnBattleAreaDigimon(card);
+                    return cardSource.IsDigimon
+                        && cardSource.HasLevel && cardSource.Level <= 4
+                        && (cardSource.CardColors.Contains(CardColor.Yellow) || cardSource.CardColors.Contains(CardColor.Black) || cardSource.HasCSTraits)
+                        && CardEffectCommons.CanPlayAsNewPermanent(cardSource, false, activateClass);
                 }
 
-                IEnumerator ActivateCoroutine(Hashtable hashtable)
+                // Using this to cover both non ESS & ESS case.
+                var thisPermament = card.PermanentOfThisCard().TopCard.PermanentOfThisCard();
```
hooperk: While we're making changes to this code, `.TopCard.PermanentOfThisCard()` is redundant, we already have the permanent form the first `card.PermanentOfThisCard()`

hooperk (changes requested):

CardEffect/BT23/Yellow/BT23_032.cs
x89rt2: I'm going to have to ask you to do that for me because I don't understand what you mean.

hooperk: `IEnumerator SharedActivateCoroutine(Hashtable hashtable, ActivateClass activateClass, bool mainText)`

Calling  `IEnumerator SharedActivateCoroutine(Hashtable hashtable, ActivateClass activateClass)` with 3 arguments will error because it only expects 2

x89rt2: Used BT23-023 Whamon as reference.

CardEffect/BT23/Yellow/BT23_032.cs
```diff
@@ -262,89 +262,93 @@ IEnumerator SelectPermanentCoroutine(Permanent permanent)
...
+                    return cardSource.IsDigimon
+                        && cardSource.HasLevel && cardSource.Level <= 4
+                        && (cardSource.CardColors.Contains(CardColor.Yellow) || cardSource.CardColors.Contains(CardColor.Black) || cardSource.HasCSTraits)
+                        && CardEffectCommons.CanPlayAsNewPermanent(cardSource, false, activateClass);
                 }
 
-                IEnumerator ActivateCoroutine(Hashtable hashtable)
+                Permanent thisPermament = null;
+                if (IsTopCard) thisPermament = card.PermanentOfThisCard();
+                else thisPermament = card.PermanentOfThisCard().TopCard.PermanentOfThisCard();
```
hooperk: We don't have to make use of isTopCard, and I'd be interested to see what happens if we don't. Especially since this code is redundant.
We need it in the method signature just so we can pass a different value, but then we can just do nothing with it and see if just having it is enough to fix the sharing OPT issue.

x89rt2: sure, we are already messing with this card so why not, let me know if I didn't do it right.

hooperk (approved):

PR #875, opened 2025-12-24
Future proof: Animal and Sea Animal

hooperk:
Not needed for anything specific right now, but it occurred to me that if we ever get a weird sea animal + animal, it would be denied by our current way of checking that a trait is one but not another. Made a different method of checking this


hooperk:
Not needed right now so going to close this

PR #876, opened 2025-12-24
Highlight the target of optional effect for BT21-083

hooperk:


CardEffect/BT21/Red/BT21_083.cs
```diff
@@ -154,8 +154,14 @@ bool CanActivateCondition(Hashtable hashtable)
 
                 bool PermanentCondition(Permanent permanent)
                 {
-                    return CardEffectCommons.IsPermanentExistsOnOwnerBattleAreaDigimon(permanent, card) &&
-                           (permanent.TopCard.EqualsTraits("Xros Heart") || permanent.TopCard.EqualsTraits("Hero"));
+                    if (CardEffectCommons.IsPermanentExistsOnOwnerBattleAreaDigimon(permanent, card) &&
+                           (permanent.TopCard.EqualsTraits("Xros Heart") || permanent.TopCard.EqualsTraits("Hero")))
+                    {
+                        activateClass.SetEffectTarget(permanent.TopCard);
```
hooperk: The CanUseCondition short circuits once it finds a valid condition, so this will most recently have been set for the permanent which matched the condition

Scripts/OptionalSkill.cs
```diff
@@ -17,7 +17,7 @@ public IEnumerator SelectOptional(ICardEffect cardEffect)
         _endSelect = false;
         _useOptional = false;
 
-        string _Message = $"Will you use \"{cardEffect.EffectName}\"?";
+        string _Message = cardEffect.EffectTarget != null ? $"Will you use \"{cardEffect.EffectName}\" targetting {cardEffect.EffectTarget.Name}?" : $"Will you use \"{cardEffect.EffectName}\"?";
```
hooperk: Need to see if there is a better way to get the proper display name?

Scripts/OptionalSkill.cs
```diff
@@ -48,15 +48,32 @@ public IEnumerator SelectOptional(ICardEffect cardEffect)
...
+            {
+                if (targetPermanent.ShowingPermanentCard != null)
+                {
+                    highlightPermanents.Add(targetPermanent.ShowingPermanentCard);
+                }
+            }
+
+            if (highlightPermanents.Count > 0)
+            {
+                GManager.instance.hideCannotSelectObject.SetUpHideCannotSelectObject(highlightPermanents, false);
```
hooperk, line 87: If I understand correctly, this is the code that highlights some objects and hides over the rest. Here I am looking to highlight both the triggering permanent if it exists and each target permanent if it exists

PR #878, opened 2025-12-24
Update BT24_023.cs

x89rt2:


x89rt2:
With this change, does it perhaps not need the                     if (CardEffectCommons.HasMatchConditionOwnersPermanent(card, CanSelectPermanentCondition1)) above?

hooperk:
> With this change, does it perhaps not need the if (CardEffectCommons.HasMatchConditionOwnersPermanent(card, CanSelectPermanentCondition1)) above?

I believe in most cases, we either need that or the maxCount =Math.min..., not both. Being fully safe would be to have both

hooperk (approved):

PR #887, opened 2025-12-26
Update BT23_056.cs

x89rt2:
Fix OP/WD being optional when it's mandatory.

### DIFF
--- a/CardEffect/BT23/Black/BT23_056.cs
+++ b/CardEffect/BT23/Black/BT23_056.cs
@@ -275,11 +275,6 @@ namespace DCGO.CardEffects.BT23
                     return CardEffectCommons.IsExistOnBattleAreaDigimon(card);
                 }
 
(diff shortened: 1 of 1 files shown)


CardEffect/BT23/Black/BT23_056.cs
```diff
@@ -50,15 +50,15 @@ bool IsCsTamer(Permanent permanent)
                     && permanent.TopCard.HasCSTraits;
             }
 
+            bool CanSelectPermamentCondition(Permanent permanent)
```
hooperk: There is another conflicting CanSelectPermanentCondition defined on line 278. It checks the exact same thing, so delete the method from lines 278-281 to prevent compilation error

hooperk (changes requested):

hooperk (approved):

PR #889, opened 2025-12-26
Implement BT24-101 and allow dynamic Digivolution cost

### LINKED ISSUE #593: BT24-101 Jupitermon

hooperk:


CardEffect/BT24/Yellow/BT24_101.cs
```diff
@@ -0,0 +1,262 @@
...
+                    return card.Owner.SecurityCards.Count;
+                }
+
+                cardEffects.Add(CardEffectFactory.AddSelfDigivolutionRequirementStaticEffect(
+                    permanentCondition: PermanentCondition,
+                    digivolutionCost: 5,
+                    ignoreDigivolutionRequirement: false,
+                    card: card,
+                    condition: null,
+                    costEquation: CostEquation)
```
x89rt2, line 51: Not understanding this code. Why is there a digivolutioncost of 5 if the costequation is the true cost. Is it doing a reduction? in that case the cost equation code on line 40 should be different.

CardEffect/BT24/Yellow/BT24_101.cs
```diff
@@ -0,0 +1,262 @@
...
+            }
+
+            bool SharedPermanentCondition(Permanent permanent)
+            {
+                return CardEffectCommons.IsPermanentExistsOnOpponentBattleAreaDigimon(permanent, card);
+            }
+
+            IEnumerator SharedActivateCoroutine(Hashtable hashtable, ActivateClass activateClass)
+            {
+                yield return ContinuousController.instance.StartCoroutine(new IDestroySecurity(
```
x89rt2, line 75: maybe add an if sec count =>1 here? or is the controller smart enough to do nothing if there are 0?

CardEffect/BT24/Yellow/BT24_101.cs
```diff
@@ -0,0 +1,262 @@
...
+                    yield return ContinuousController.instance.StartCoroutine(CardEffectCommons.ChangeDigimonDP(
+                        targetPermanent: permanent,
+                        changeValue: -13000,
+                        effectDuration: EffectDuration.UntilOpponentTurnEnd,
+                        activateClass: activateClass));
+                }
+
+                if(card.Owner.SecurityCards.Count <= 1)
+                {
+                    yield return ContinuousController.instance.StartCoroutine(new IRecovery(card.Owner, 2, activateClass).Recovery());
```
x89rt2, line 115: I have to assume this doesn't need a if deck count => 1, correct?

CardEffect/BT24/Yellow/BT24_101.cs
```diff
@@ -0,0 +1,262 @@
...
+
+            #endregion
+
+            #region All Turns When Sec Removed
+
+            if (timing == EffectTiming.OnLoseSecurity)
+            {
+                ActivateClass activateClass = new ActivateClass();
+                activateClass.SetUpICardEffect("Trash Opponent's top security", CanUseCondition, card);
+                activateClass.SetUpActivateClass(CanActivateCondition, ActivateCoroutine, 1, true, EffectDiscription());
```
x89rt2: false

CardEffect/BT24/Yellow/BT24_101.cs
```diff
@@ -0,0 +1,262 @@
...
+
+                bool CanUseCondition(Hashtable hashtable)
+                {
+                    return CardEffectCommons.IsExistOnBattleAreaDigimon(card)
+                        && CardEffectCommons.CanTriggerWhenLoseSecurity(hashtable, (player) => player == card.Owner);
+                }
+
+                bool CanActivateCondition(Hashtable hashtable)
+                {
+                    return CardEffectCommons.IsExistOnBattleAreaDigimon(card);
```
x89rt2, line 180: do we maybe want a check on if oppo sec => 1?

CardEffect/BT24/Yellow/BT24_101.cs
```diff
@@ -0,0 +1,262 @@
...
+                }
+
+                bool CanSelectPermanentCondition(Permanent permanent)
+                {
+                    return CardEffectCommons.IsPermanentExistsOnBattleArea(permanent)
+                        && permanent.IsDigimon;
+                }
```
x89rt2: unnecessary

CardEffect/BT24/Yellow/BT24_101.cs
```diff
@@ -0,0 +1,262 @@
...
+            }
+
+            #endregion
+
+            #region All Turns Protect TS
+
+            if (timing == EffectTiming.WhenRemoveField)
+            {
+                ActivateClass activateClass = new ActivateClass();
+                activateClass.SetUpICardEffect("By suspending this digimon, your [TS]/[Aqua]/[Sea Animal] digimon wont leave the field", CanUseCondition, card);
```
x89rt2: incorrect description

CardEffect/BT24/Yellow/BT24_101.cs
```diff
@@ -0,0 +1,262 @@
...
+                        && permanent.TopCard.HasTSTraits;
+                }
+
+                IEnumerator ActivateCoroutine(Hashtable hashtable)
+                {
+                    yield return ContinuousController.instance.StartCoroutine(new IDestroySecurity(
+                        player: card.Owner.Enemy,
+                        destroySecurityCount: 1,
+                        cardEffect: activateClass,
+                        fromTop: true).DestroySecurity());
```
x89rt2, line 238: We could probably use a success confirmation here? I don't believe an effect exists that would prevent sec removal, but this feels more realistic then some other hypotheticals...

CardEffect/BT24/Yellow/BT24_101.cs
```diff
@@ -0,0 +1,262 @@
...
+
+                    List<Permanent> protectedPermanents = CardEffectCommons.GetPermanentsFromHashtable(hashtable)
+                                .Filter(PermanentCondition);
+
+                    foreach (Permanent permanent in protectedPermanents)
+                    {
+                        permanent.willBeRemoveField = false;
+                        permanent.HideDeleteEffect();
+                        permanent.HideHandBounceEffect();
+                        permanent.HideDeckBounceEffect();
```
x89rt2: might need more things, refer to Venus comment.

x89rt2 (changes requested):

CardEffect/BT24/Yellow/BT24_101.cs
x89rt2: Ah, this might be related to your changes in the other document. Do as you will as that's over my head.

hooperk: cost is a required parameter so I have to put something in for it. Left 5 rather than bothered changing it. The cost gets ignored in favor of the CostEquation when that is defined

CardEffect/BT24/Yellow/BT24_101.cs
hooperk: IDestroySecurity.DestroySecurity() checks for that

CardEffect/BT24/Yellow/BT24_101.cs
hooperk: Yup, IRecovery.Recovery() checks it can do so

CardEffect/BT24/Yellow/BT24_101.cs
hooperk: The IDestroySecurity.DestroySecurity checks anyway, and it is technically allowed to whiff this effect

CardEffect/BT24/Yellow/BT24_101.cs
hooperk: I couldn't find a way to confirm the success but I'll think on it

CardEffect/BT24/Yellow/BT24_101.cs
x89rt2: I'm looking at some other examples and the only one I'm seeing that's not here is:

selectedPermanent.HideWillRemoveFieldEffect();

CardEffect/BT24/Yellow/BT24_101.cs
```diff
@@ -0,0 +1,262 @@
...
+
+                        foreach (Permanent permanent in protectedPermanents)
+                        {
+                            permanent.willBeRemoveField = false;
+                            permanent.HideDeleteEffect();
+                            permanent.HideHandBounceEffect();
+                            permanent.HideDeckBounceEffect();
+                            permanent.HideWillRemoveFieldEffect();
+                        }
+                    }
```
Michael8818, line 253: best way to confirm trashing is instead of tracking security count track the top security card at time of activation, then check if that card is either in trash. There isn't anything currently that would make this relevant at the current moment, but to consider later.

Michael8818 (approved):

CardEffect/BT24/Yellow/BT24_101.cs
hooperk: Nice, I'll keep that in mind

PR #890, opened 2025-12-26
Implement BT24-099

### LINKED ISSUE #591: BT24-099
Image

hooperk:


CardEffect/BT24/Purple/BT24_099.cs
```diff
@@ -0,0 +1,289 @@
...
+            #region All Turns - Delay
+
+            if(timing == EffectTiming.OnDestroyedAnyone)
+            {
+                ActivateClass activateClass = new ActivateClass();
+                activateClass.SetUpICardEffect("Link 1 [Appmon] from trash", CanUseCondition, card);
+                activateClass.SetUpActivateClass(null, ActivateCoroutine, -1, true, EffectDescription());
+                cardEffects.Add(activateClass);
+
+                string EffectDescription() => "[All Turns] When Digimon are deleted, <Delay>.\r\n• You may link 1 [Appmon] trait Digimon card from your trash to 1 of your Digimon without paying the cost.";
```
x89rt2: Get rid of the .\r\n•.

CardEffect/BT24/Purple/BT24_099.cs
```diff
@@ -0,0 +1,289 @@
...
+                bool CanUseCondition(Hashtable hashtable)
+                {
+                    return CardEffectCommons.IsExistOnBattleArea(card)
+                        && CardEffectCommons.CanDeclareOptionDelayEffect(card)
+                        && CardEffectCommons.CanTriggerOnPermanentDeleted(hashtable, PermanentCondition);
+                }
+
+                bool PermamentCondition(Permanent permanent)
+                {
+                    return permanent.TopCard.IsDigimon;
```
x89rt2: what's the realistic example of a difference between this and permanent.IsDigimon?

CardEffect/BT24/Purple/BT24_099.cs
```diff
@@ -0,0 +1,289 @@
...
+                }
+
+                bool CanActivateCondition(Hashtable hashtable)
+                {
+                    return CardEffectCommons.IsExistOnBattleArea(card)
+                        && CardEffectCommons.HasMatchConditionOwnersCardInTrash(card, CardCondition);
```
x89rt2: Remove the matchcondition on canactivate. Sometimes you just want to pop the delay with nothing in trash to avoid justimon popping it for you for their benefit.

x89rt2: That also means you don't need the cardcondition below this I believe.

CardEffect/BT24/Purple/BT24_099.cs
```diff
@@ -0,0 +1,289 @@
...
+                bool OwnPermamentCondition(Permanent permanent) => CardEffectCommons.IsPermanentExistsOnOwnerBattleAreaDigimon(permanent, card);
+                bool LinkCardCondition(CardSource cardSource, Permanent targetPermament) => cardSource.IsDigimon && cardSource.HasAppmonTraits && cardSource.CanLinkToTargetPermanent(targetPermament, false);
+
+                IEnumerator ActivateCoroutine(Hashtable hashtable)
+                {
+                    bool delaySuccessful = false;
+
+                    yield return ContinuousController.instance.StartCoroutine(CardEffectCommons.DeletePeremanentAndProcessAccordingToResult(targetPermanents: new List<Permanent>() { card.PermanentOfThisCard() }, activateClass: activateClass, successProcess: permanents => SuccessProcess(), failureProcess: null));
+
+                    IEnumerator SuccessProcess()
```
x89rt2, line 159: I don't know if you've copied this from some other card but if you declare to activate delay, there is currently no way to get it to NOT delete itself so I don't think you need a success check here.

x89rt2 (changes requested):
One other thing, I think it would be better to choose the trash link card first and THEN choose what to link it to on field. If we do it in the current order, you'll have to look at trash mid effect to properly plan out what to link to which of your digimon, when you are more likely to instinctively know what is in your field over what is in your trash.

CardEffect/BT24/Purple/BT24_099.cs
hooperk: Did we not want that? I copied from one of yours

hooperk:
> One other thing, I think it would be better to choose the trash link card first and THEN choose what to link it to on field. If we do it in the current order, you'll have to look at trash mid effect to properly plan out what to link to which of your digimon, when you are more likely to instinctively know what is in your field over what is in your trash.

That creates some problems with the Can Link to Permanent check, although I suppose the logic could just be reversed and Appmon are fairly non-restrictive anyway so far. I did make this to match how other similar cards do it, like Heart of Music

CardEffect/BT24/Purple/BT24_099.cs
x89rt2: Not sure? I assume this is trash code from when I copied from card dev. Do you know if it does anything?

CardEffect/BT24/Purple/BT24_099.cs
hooperk: This is the standard way we do every Delay, and in case something later on breaks that it's a good pattern to have.

CardEffect/BT24/Purple/BT24_099.cs
hooperk: I just forgot permanent has it as well

x89rt2:
> > One other thing, I think it would be better to choose the trash link card first and THEN choose what to link it to on field. If we do it in the current order, you'll have to look at trash mid effect to properly plan out what to link to which of your digimon, when you are more likely to instinctively know what is in your field over what is in your trash.
> 
> That creates some problems with the Can Link to Permanent check, although I suppose the logic could just be reversed and Appmon are fairly non-restrictive anyway so far. I did make this to match how other similar cards do it, like Heart of Music

That one is from hand no? Since it's hand, there's no real menus obstructing the battle area so you can make informed decisions without clicking anywhere with that one.

CardEffect/BT24/Purple/BT24_099.cs
hooperk: It creates a new line and has the circle symbol. So basically it should actually look like this wherever it is displayed, including the second part being on a different line:
[All Turns] When Digimon are deleted, <Delay>.
• You may link 1 [Appmon] trait Digimon card from your trash to 1 of your Digimon without paying the cost.

x89rt2: well, do as you will

CardEffect/BT24/Purple/BT24_099.cs
```diff
@@ -0,0 +1,282 @@
...
+                    return permanent.IsDigimon;
+                }
+
+                bool CanActivateCondition(Hashtable hashtable)
+                {
+                    return CardEffectCommons.IsExistOnBattleArea(card);
+                }
+
+                bool OwnPermamentCondition(CardSource cardsource, Permanent permanent) => CardEffectCommons.IsPermanentExistsOnOwnerBattleAreaDigimon(permanent, card) && cardSource.CanLinkToTargetPermanent(permanent, false);
+                bool LinkCardCondition(CardSource cardSource) => cardSource.IsDigimon && cardSource.HasAppmonTraits;
```
x89rt2: I've got the basics, but are we sure they won't be able to select an appmon in trash with no link condition like gaiamon and attempt to link with it here?

hooperk: You're right. I need to look for a way to check for a link condition, or swap it back around so we pick the permanent first. I'll look tomorrow

CardEffect/BT24/Purple/BT24_099.cs
```diff
@@ -0,0 +1,282 @@
...
+                    return permanent.IsDigimon;
+                }
+
+                bool CanActivateCondition(Hashtable hashtable)
+                {
+                    return CardEffectCommons.IsExistOnBattleArea(card);
+                }
+
+                bool OwnPermamentCondition(CardSource cardsource, Permanent permanent) => CardEffectCommons.IsPermanentExistsOnOwnerBattleAreaDigimon(permanent, card) && cardSource.CanLinkToTargetPermanent(permanent, false);
+                bool LinkCardCondition(CardSource cardSource) => cardSource.IsDigimon && cardSource.HasAppmonTraits && cardSource.CanLink(false);
```
x89rt2, line 151: Hm... It's sad, but I think with maquina coming up, this doesn't work. Correct me if I'm wrong of course. I think we will have to go back to the method you did originally with selecting a card to get linked to first and then going to trash to pick a card to link that can link to that first choice.

hooperk: It won't be broken then because of the HasAppmonTrait... but yeah when it is this way around we cannot guarantee it won't introduce a bug down the line

Michael8818 (approved):

Michael8818 (approved):

PR #907, opened 2025-12-27
Fix all BT24 Discription misspelling to Description and update file name of BT24_092 to add .cs

x89rt2:


x89rt2:
Used replace all command so there should be 0 mismatch.

hooperk (approved):

PR #1013, opened 2025-12-28
Allow user to discard all remaining optional effects

hooperk:
Provides the option to choose not to activate all remaining optional effects when only optional effects remain to be chosen from.

Useful for certain decks that have a lot of identical optional effects trigger at once but only need to activate 1, like Xros Heart or Eaters.

https://canary.discord.com/channels/839041812932329512/1454784427019210887/1454784427019210887

### DIFF
--- a/Scripts/MultipleSkills.cs
+++ b/Scripts/MultipleSkills.cs
@@ -266,13 +266,17 @@ public class MultipleSkills : MonoBehaviourPunCallbacks
                             {
                                 yield return StartCoroutine(GManager.instance.selectCardPanel.OpenSelectCardPanel(
                                 Message: "Multiple effects are triggered.\nChoose which effect to process.",
+                                NotSelectButtonMessage: "Don't activate these effects.",
+                                EndSelectButtonMessage: "End Selection", 
+                                _OnClickNotSelectButtonAction: null, 
+                                _OnClickEndSelectButtonAction: null,
                                 RootCardSources: RootCardSources,
                                 _CanTargetCondition: (cardSource) => true,
                                 _CanTargetCondition_ByPreSelecetedList: null,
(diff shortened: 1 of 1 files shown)


x89rt2:
Hm... I don't know how I feel about this one. I WOULD like to get an option that auto chooses effect orders for that chain of effects when multiple effects trigger at once, but I don't know if this conflicts with that.

hooperk:
> Hm... I don't know how I feel about this one. I WOULD like to get an option that auto chooses effect orders for that chain of effects when multiple effects trigger at once, but I don't know if this conflicts with that.

I believe we have something like that, and I'd like this to be as well as it not instead of it. This is ONLY for the case where all remaining effects are optional, and should be essentially the same as just saying no to each of those in turn, but with much less clicks than even auto-ordering would be (you'd still need to click no for each one then)

x89rt2:
> > Hm... I don't know how I feel about this one. I WOULD like to get an option that auto chooses effect orders for that chain of effects when multiple effects trigger at once, but I don't know if this conflicts with that.
> 
> I believe we have something like that, and I'd like this to be as well as it not instead of it. This is ONLY for the case where all remaining effects are optional, and should be essentially the same as just saying no to each of those in turn, but with much less clicks than even auto-ordering would be (you'd still need to click no for each one then)

we sort of do, it's an option in the config, but I would love to somehow someday get it as a button on selection screens like this: 
https://cdn.discordapp.com/attachments/1387963655928152217/1415124743727747223/Untitled.png?ex=695a565e&is=695904de&hm=dcdae50e2d57eb41a7fe44998a527eea764c663cbd31b3b555d9529c836996aa&

I feel like that might be outside of cardeffects though

PR #1017, opened 2025-12-29
Update BT9_018.cs

x89rt2:
Fix WD not activating if no digimon in battle area and weird symbols.

### DIFF
--- a/CardEffect/BT9/Red/BT9_018.cs
+++ b/CardEffect/BT9/Red/BT9_018.cs
@@ -21,7 +21,7 @@ public class BT9_018 : CEntity_Effect
 
             string EffectDiscription()
             {
-                return "[When Digivolving] For each Tamer your opponent has in play, suspend 1 of your opponentf Digimon and gain 1 memory.";
+                return "[When Digivolving] For each Tamer your opponent has in play, suspend 1 of your opponent's Digimon and gain 1 memory.";
             }
 
             bool CanSelectPermanentCondition(Permanent permanent)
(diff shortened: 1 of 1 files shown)


CardEffect/BT9/Red/BT9_018.cs
```diff
@@ -98,7 +90,7 @@ IEnumerator ActivateCoroutine(Hashtable _hashtable)
 
             string EffectDiscription()
             {
-                return "[All Turns][Once Per Turn] When an opponent�f Digimon with 6000 DP or less becomes suspended, you may delete that Digimon.";
+                return "[All Turns][Once Per Turn] When an opponentf Digimon with 6000 DP or less becomes suspended, you may delete that Digimon.";
```
hooperk: Looks like a symbol was replaced here. Could we replace opponentf with "opponent's"?

CardEffect/BT9/Red/BT9_018.cs
```diff
@@ -21,7 +21,7 @@ public override List<ICardEffect> CardEffects(EffectTiming timing, CardSource ca
 
             string EffectDiscription()
             {
-                return "[When Digivolving] For each Tamer your opponent has in play, suspend 1 of your opponent�f Digimon and gain 1 memory.";
+                return "[When Digivolving] For each Tamer your opponent has in play, suspend 1 of your opponentf Digimon and gain 1 memory.";
```
hooperk: Same here, opponentf => "opponent's"

hooperk (changes requested):

hooperk (approved):

PR #1018, opened 2025-12-29
Implement BT24-065

### LINKED ISSUE #557: BT24-065 Diaboromon (X Antibody)

hooperk:


CardEffect/BT24/Black/BT24_065.cs
```diff
@@ -0,0 +1,270 @@
...
+                cardEffects.Add(CardEffectFactory.BlockerSelfStaticEffect(isInheritedEffect: false, card: card, condition: null));
+            }
+            #endregion
+
+            #region When Digivolving
+
+            if (timing == EffectTiming.OnEnterFieldAnyone)
+            {
+                ActivateClass activateClass = new ActivateClass();
+                activateClass.SetUpICardEffect("De-digivolve 1 opponent's Digimon. Delete all their highest play cost Digimon.", CanUseCondition, card);
```
x89rt2: Not the most accurate description, maybe remove the one to make the singular or plural ambigues like: "De-digivolve opponent's Digimon. Delete all their highest play cost Digimon."

CardEffect/BT24/Black/BT24_065.cs
```diff
@@ -0,0 +1,270 @@
...
+        {
+            List<ICardEffect> cardEffects = new List<ICardEffect>();
+
+            #region Alternate Digivolution
+
+            if (timing == EffectTiming.None)
+            {
+                bool PermanentCondition(Permanent targetPermanent)
+                {
+                    return targetPermanent.TopCard.EqualsCardName("Diaboromon");
```
x89rt2, line 21: Do we need to worry about trying to digivolve over tokens here?

CardEffect/BT24/Black/BT24_065.cs
```diff
@@ -0,0 +1,270 @@
...
+                }
+
+                bool CanSelectPermanentCondition(Permanent permanent)
+                {
+                    return CardEffectCommons.IsPermanentExistsOnOpponentBattleAreaDigimon(permanent);
+                }
+
+                bool CanDeleteCondition(Permanent permanent)
+                {
+                    return CardEffectCommons.IsMinCost(permanent, card.Owner.Enemy, true);
```
x89rt2: It's max cost. Does the true make it max or something?

CardEffect/BT24/Black/BT24_065.cs
```diff
@@ -0,0 +1,270 @@
...
+                    selectPermanentEffect.SetUpCustomMessage($"Select 1 digimon to <De-Digivolve 1> {digimonCount} times", "Opponent is selecting a Digimon to De-Digivolve.");
+                    selectPermanentEffect.SetUpCustomMessage_ShowCard("Played Card");
+
+                    yield return ContinuousController.instance.StartCoroutine(selectPermanentEffect.Activate());
+
+                    IEnumerator SelectPermanentCoroutine(Permanent permanent)
+                    {
+                        if (permanent != null)
+                        {
+                            for (int i = 0; i < digimonCount; i++)
```
x89rt2, line 104: could do the following, copying the likes of lordknight x to simplify this:

int maxCount1 = card.Owner.GetBattleAreaDigimons().Count;

IEnumerator DeDigivolvePermanent(Permanent permanent)
                    {
                        yield return ContinuousController.instance.StartCoroutine(new IDegeneration(permanent, maxCount1, activateClass).Degeneration());
                    }

x89rt2 (changes requested):

CardEffect/BT24/Black/BT24_065.cs
hooperk: No, there is a general check for that

CardEffect/BT24/Black/BT24_065.cs
hooperk: Degenerate 2 and degenerate 1 twice are not the same. This is actually a bug on LordKnightmon X.

The biggest case it makes a difference is becoming something with immunity partway through, which is niche enough to not come up often.

I copied my code from one of the Machinedramons

x89rt2: Let me see if I can get someone to verify lkx is bugged then.

hooperk: Typical way to test is to have a suspended digimon with BT15-047 Kabuterimon as the second digimon down, and try something that would de-digivolve past it.
Expected: It stops at Kabuterimon

x89rt2: I'm familiar, just really surprised this hasn't come up before if this really was an issue. I feel like this is something I would have probably tested at some point a while ago.

hooperk: Unfortunately, I hope it is a bug with LordKnightmon, because if not then it is a bug with De-Digivolve. The command LordKnightmon has right now should be De-Digivolve X, unless there is some different method to use for de-digivolve X, or some boolean to pass to tell it whether it acts like De-Digivolve 1 X times or De-Digivolve X.

x89rt2 (approved):

PR #1019, opened 2025-12-29
Implement BT24-079

### LINKED ISSUE #571: BT24-079
Image

hooperk:


CardEffect/BT24/Purple/BT24_079.cs
```diff
@@ -0,0 +1,393 @@
...
+            if (timing == EffectTiming.None) cardEffects.Add(CardEffectFactory.ChangeSelfLinkMaxStaticEffect(1, false, card, null));
+            #endregion
+
+            #region When Digivolving
+
+            if (timing == EffectTiming.OnEnterFieldAnyone)
+            {
+                ActivateClass activateClass = new ActivateClass();
+                activateClass.SetUpICardEffect("Play a lvl 4- from trash. Add new link to a digimon", CanUseCondition, card);
+                activateClass.SetUpActivateClass(CanActivateCondition, ActivateCoroutine, -1, false, EffectDiscription());
```
x89rt2, line 51: since the entire effect is optional, we should probably make this true for an easy 1 click don't do anything.

x89rt2: unless you think it complicates the AT

x89rt2 (changes requested):

CardEffect/BT24/Purple/BT24_079.cs
hooperk: It was more a case that technically the effect is not optional, it has 2 optional subclauses. And to avoid the potential accidental say no when you do want the second part

x89rt2: good call, yea, let's leave as is.

x89rt2 (approved):

Issue #388, opened 2025-11-27
BT7-091 Koichi Kimura - Bug #2


LewisWelch94:
ESS On Deletion is behaving as this card's On Deletion instead of top card's. - Probably explains above bug
https://discord.com/channels/1095717982522585098/1229224709007343687/1407058425296191618

PR #1028, opened 2025-12-29
Update EX10-037

### LINKED ISSUE #358: EX10-037 Impmon
When trashed effect is bypassing gankoo X immunity
https://discord.com/channels/1095717982522585098/1436168988106166393

hooperk:
Mark the Trash effect as a Digimon Effect, so that Cannot be effected by Digimon Effect digimon are immune.
I believe since this is not a permanent for the trash effect, the code which automatically sets it to be a Digimon effect is not setting this.

### DIFF
--- a/CardEffect/EX10/Purple/EX10_037.cs
+++ b/CardEffect/EX10/Purple/EX10_037.cs
@@ -16,8 +16,9 @@ namespace DCGO.CardEffects.EX10
             if (timing == EffectTiming.OnDiscardLibrary)
             {
                 ActivateClass activateClass = new ActivateClass();
-                activateClass.SetUpICardEffect("delete 1 level 4 or lower digimon", CanUseCondition, card);
+                activateClass.SetUpICardEffect("Delete 1 level 4 or lower digimon", CanUseCondition, card);
                 activateClass.SetUpActivateClass(CanActivateCondition, ActivateCoroutine, -1, false, EffectDiscription());
+                activateClass.SetIsDigimonEffect(true);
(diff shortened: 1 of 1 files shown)


x89rt2:
@Michael8818 this one might explain the ones in the BT24 cards. Although, to be fair, I don't think the BT24 ones have any effect currently in the game that would make theirs matter. More for future proofing I guess?

Michael8818 (approved):

x89rt2:
@hooperk if this ends up fixing this issue, is this something you can also do for tamers and options?

hooperk:
> @hooperk if this ends up fixing this issue, is this something you can also do for tamers and options?

@x89rt2 there is a SetIsTamerEffect that could be set the same way, although there are currently not any cards that do a tamer effect without existing as a permanent so we shouldn't need to. There is no equivalent bool IsOptionEffect, but I don't know of anything specifying option effects currently to have an issue with that

x89rt2:
@hooperk my concern was https://github.com/DCGO2/DCGO-Card-Scripts/issues/388 although reading it again it might be the opposite? not sure, morning brain right now. Similarly, whe have the 3M options when trashed effects.

hooperk:
> @hooperk my concern was #388 although reading it again it might be the opposite? not sure, morning brain right now. Similarly, whe have the 3M options when trashed effects.

Hmm, Koichi already has this code to attempt to count as a Digimon effect. I actually think the bug may be that the memory blockers that prevent memory gain don't use these IsDigimonEffect/IsTamerEffect to make their decision

PR #1053, opened 2026-01-01
Update EX10-061 not considered removing from security

hooperk:


CardEffect/EX10/White/EX10_061.cs
```diff
@@ -219,6 +219,9 @@ IEnumerator SelectCardCoroutine(CardSource cardSource)
                             if (selectedCards.Count >= 1)
                             {
                                 GManager.instance.GetComponent<SelectDigiXrosClass>().AddDigivolutionCardInfos(new AddDigivolutionCardsInfo(activateClass, selectedCards));
+                                yield return ContinuousController.instance.StartCoroutine(new IReduceSecurity(
+                                    player: card.Owner,
+                                    refSkillInfos: ref ContinuousController.instance.nullSkillInfos).ReduceSecurity());
```
x89rt2: really? that easy? this doesn't actually trash a sec or anything correct?

hooperk: Nope, this tells the game to trigger security reduced effects. The reporter mentioned that WaruMonzaemon works fine, that is because within the code that plays a permanent from a root, if that root is security it automatically calls this method. However, the add to digivolution cards does not have code to do that, so we have to do it manually. You can see similar on add from security to hand cards like BT1_087 TK. The SelectCardEffect does not know to automatically trigger this, so TK's effect has to specifically call it

x89rt2 (approved):

PR #1054, opened 2026-01-01
Update BT24-076

hooperk:


CardEffect/BT24/Purple/BT24_076.cs
```diff
@@ -18,6 +18,7 @@ public override List<ICardEffect> CardEffects(EffectTiming timing, CardSource ca
                 ActivateClass activateClass = new ActivateClass();
                 activateClass.SetUpICardEffect("Play this card from trash with reduced cost", CanUseCondition, card);
                 activateClass.SetUpActivateClass(null, ActivateCoroutine, -1, false, EffectDescription());
+                activateClass.SetIsDigimonEffect(true);
```
Michael8818, line 21: What is the point of this change?

hooperk: As a trash effect, this isn't automatically set. I noticed other trash effects are setting this.
It is in case of a future "cannot play a card by Digimon effect"

Michael8818 (approved):

PR #1057, opened 2026-01-02
Fix BT18-033 if Patamon fails to play, should be trashed instead

hooperk:
image

### DIFF
--- a/CardEffect/BT18/Yellow/BT18_033.cs
+++ b/CardEffect/BT18/Yellow/BT18_033.cs
@@ -121,6 +121,11 @@ namespace DCGO.CardEffects.BT18
                           root: SelectCardEffect.Root.Hand,
                           activateETB: true,
                           isBreedingArea: true));
+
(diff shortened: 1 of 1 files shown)


CardEffect/BT18/Yellow/BT18_033.cs
```diff
@@ -121,6 +121,11 @@ IEnumerator AfterSelectCardCoroutine(List<CardSource> cardSources)
                           root: SelectCardEffect.Root.Hand,
                           activateETB: true,
                           isBreedingArea: true));
+
+                        if (!card.Owner.GetBreedingAreaPermanents().Contains(card.PermanentOfThisCard()))
+                        {
+                            yield return ContinuousController.instance.StartCoroutine(CardObjectController.AddTrashCard(card));
+                        }
```
x89rt2, line 128: Does the sim considered this trashed and activate purple cards with this code?

hooperk: No, this is the same code used by options to just go to trash when they have finished resolving.

Michael8818 (approved):

PR #1061, opened 2026-01-02
Update BT18_092.cs

x89rt2:
Too annoying to have to keep hitting multiple buttons


hooperk (approved):

hooperk:
I agree, even if the effect is optional, if it immediately opens a dialogue with no select is true and isn't once per turn I feel it better to just let the user no select instead of asking them

PR #1069, opened 2026-01-02
Implement EX11_066 card effect functionality

x89rt2:
@hooperk something like this maybe?


hooperk:
> @hooperk something like this maybe?

Almost, but currently if you have 1 vemmon you can't finish selecting, and if you have 2 this is going to allow you to tuck them under different digimon. But I'll take this and look to fix those parts.
Edit: It is actually smart about the Max Count

PR #1070, opened 2026-01-03
Ensure inherited and link effect treated as Digimon Effect

### LINKED ISSUE #388: BT7-091 Koichi Kimura - Bug #2

hooperk:
For Inherited On Deletion effect, since the card no longer has a permanent it is being treated as a tamer effect again


Scripts/AutoProcessing.cs
```diff
@@ -61,21 +61,27 @@ public void PutStackedSkill(SkillInfo skillInfo)
...
             if (permanent.IsDigimon)
             {
-                skillInfo.CardEffect.SetIsDigimonEffect(true);
+                cardEffect.SetIsDigimonEffect(true);
             }
         }
 
+        if (cardEffect.IsInheritedEffect || cardEffect.IsLinkedEffect)
+        {
+            cardEffect.SetIsDigimonEffect(true);
```
hooperk, line 79: Inherited and Link effects can only be Digimon effects. While Link has no current issues, it is possible an option will add an OnDeletion link effect soon

Scripts/AutoProcessing.cs
```diff
@@ -84,22 +90,22 @@ public void PutStackedSkill(SkillInfo skillInfo)
         {
             if (permanent.IsTamer)
             {
-                skillInfo.CardEffect.SetIsTamerEffect(true);
+                cardEffect.SetIsTamerEffect(true);
             }
         }
 
-        else if (card.IsTamer)
+        else if (card.IsTamer && !cardEffect.IsInheritedEffect)
```
hooperk, line 97: Don't set the effect to be a tamer effect if it is inherited

hooperk:
@x89rt2 Pretty sure this is the cause of the koichi bug. For the on deletion, it is no longer a permanent so the code was treating it as a tamer card instead. I've adjusted it so inherited effects are only ever Digimon effects

Scripts/AutoProcessing.cs
hooperk: Bt24 Docmon necessitates this for link

x89rt2: Maybe update the PR name?

hooperk:
Did not work

PR #1079, opened 2026-01-05
Update BT24-051

hooperk:


CardEffect/BT24/Green/BT24_051.cs
```diff
@@ -403,12 +404,35 @@ bool PermanentCondition(Permanent permanent)
                     card: card,
                     condition: CanUseCondition));
```
x89rt2, line 405: Is this rush one fine?

hooperk: Yeah. There exists a "RushStaticEffect" that gives all permanents that matches the PermanentCondition Rush. I had assumed there was a pierce equivalent, but there is only a factory method to give yourself or exactly 1 other permanent pierce. Had to dig out BT2 Taiga for this code

x89rt2 (approved):

PR #1082, opened 2026-01-05
Update BT24-018 missign digivolution requirements

hooperk:


x89rt2 (approved):

x89rt2:
there is supposedly a bug on trashing security, need more info on that one but might want to do it here as well once discovered.

Michael8818 (approved):

PR #1089, opened 2026-01-05
Refactor hashtable usage in ST22_06.cs

x89rt2:
Only difference I see between P-187 Maste and this.


Michael8818 (approved):
This is correct, the reason for this is the hashtable being used originally was the hashtable data that triggered the all turns in the first place. Playing an option card. SO when you go to put a digimon in security the checks are reading the "you used an option card" as what is putting said card into security. and would make it seem like an option is placing the card into security not a digimon effect. Which would bypass a lot of immunities.

PR #1110, opened 2026-01-06
Update BT24_006.cs

x89rt2:
Fix description and modify CanUse slightly to see if it triggers now.


x89rt2:
Not sure the effect needs fixing, but the description should. Waiting on extra confirmation on if ESS is triggering or not.

x89rt2:
No issue so far so I'll be reverting the code changes tomorrow to only keep the text change.

PR #1114, opened 2026-01-06
Update BT24_025.cs

x89rt2:
We have a problem here where currently it's evoing for no cost. I'm looking into old cards to see if we have anything similar. Nothing similar, not sure how to fix.


x89rt2:
More complex, someone else will handle it.

PR #1119, opened 2026-01-06
Implement EX11-004 with new EffectTiming type

### LINKED ISSUE #745: EX11-004 - Kapurimon
Image

hooperk:


Scripts/CardController.cs
```diff
@@ -4859,6 +4859,55 @@ public IEnumerator AddSecurity()
         yield return ContinuousController.instance.StartCoroutine(GManager.instance.autoProcessing.StackSkillInfos(hashtable, EffectTiming.OnAddSecurity));
 
         #endregion
+
+        #region "When face up cards are added"
+        if (!_cardSource.IsFlipped)
+        {
+            yield return ContinuousController.instance.StartCoroutine(GManager.instance.autoProcessing.StackSkillInfos(hashtable, EffectTiming.OnFaceUpSecurityIncreased));
```
hooperk: Conveniently the AddSecurity Hashtable is the same as what we want. But want to double check if it is safe to use the same for both or if something might alter one and it best I make another?

Michael8818 (changes requested):
Problem I have here is GManager.OnSecurityStackChanged?.Invoke(card.Owner.Enemy); was removed from several cards, but I don't see it added in the new method used. Unless I am blind.

This is important because this event triggers the eyeball icon on a players security (among other things if I recall) might want to check it out.

hooperk:
@Michael8818 GManager.OnSecurityStackChanged?.Invoke is performed by CardSource.SetFace used in this method. The owner passed to the method is correctly the owner of the flipped card

Michael8818 (approved):

PR #1124, opened 2026-01-06
Rename and update Diaboromon card name check

x89rt2:


CardEffect/EX6/Black/EX6_041.cs
```diff
@@ -35,9 +35,9 @@ bool HasDiaboromonInNameField(Permanent permanent)
                     return false;
                 }
 
-                bool HasDiaboromonInNameHand(CardSource cardSource)
+                bool DiaboromonHand(CardSource cardSource)
                 {
-                    return cardSource.ContainsCardName("Diaboromon");
+                    return cardSource.EqualsCardName("Diaboromon");
```
hooperk, line 40: Since this doesn't include a check for being in hand, would something like "IsDiaboromon" suit better?

x89rt2: sure, shouldn't matter but can do

hooperk (approved):

PR #1125, opened 2026-01-06
Update BT17_100.cs

x89rt2:
Fix Main from exact name to name in text

### DIFF
--- a/CardEffect/BT17/Black/BT17_100.cs
+++ b/CardEffect/BT17/Black/BT17_100.cs
@@ -37,7 +37,7 @@ namespace DCGO.CardEffects.BT17
                         {
                             if (!permanent.IsToken)
                             {
-                                if (permanent.TopCard.EqualsCardName("Diaboromon"))
+                                if (permanent.TopCard.CardNames.Contains("Diaboromon"))
                                 {
                                     if (permanent.DigivolutionCards.Count(IsDoomsdayClock) == 0)
                                     {
--- a/CardEffect/BT17/Black/BT17_100.cs
+++ b/CardEffect/BT17/Black/BT17_100.cs
@@ -37,7 +37,7 @@ namespace DCGO.CardEffects.BT17
                         {
                             if (!permanent.IsToken)
                             {
-                                if (permanent.TopCard.CardNames.Contains("Diaboromon"))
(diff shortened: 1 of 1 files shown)


CardEffect/BT17/Black/BT17_100.cs
```diff
@@ -37,7 +37,7 @@ bool HasDiaboromonWithoutClock(Permanent permanent)
                         {
                             if (!permanent.IsToken)
                             {
-                                if (permanent.TopCard.EqualsCardName("Diaboromon"))
+                                if (permanent.TopCard.CardNames.Contains("Diaboromon"))
```
hooperk: Use ContainsCardName

hooperk (dismissed):

Michael8818 (approved):

Michael8818 (approved):

CardEffect/BT17/Black/BT17_100.cs
x89rt2: sure, but what's the difference?

hooperk: CardNames.Contains is actually close to EqualsCardName.
It is checking if any of the cardnames of this card is exactly the passed string. Using C# .equals().

ContainsCardName is our equivalent to EqualsCardName for checking if any of those card names contains the string we are passing and has extra stuff like removing spaces and comparing lower cases if needed as well. It is also the central place well make any changes if we ever need to.

x89rt2: That is VERY annoying. First impression is that contains is "in it's name".

hooperk: To give a little more explanation:
The .Contains you were using is a C# method, not one we wrote, and behaves differently depending on what it is used on. If used on a single String, it will do what you were wanting, "does that string contain this one"
However, CardNames is a '
`List<string>` and when you use .Contains on a list it becomes "does this list contain an object equal to this one, in this case does it contain an object equal to this string.

We could do something like CardNames.Some(name => name.Contains("Diaboromon"))
Which is equivalent to "does this list of strings contain a string which contains "Diaboromon".

That's just to explain though, if we stick with the special methods we made they will handle things for us, and slightly better than that example anyway

PR #1130, opened 2026-01-07
Update BT24_025.cs

x89rt2:
Fix not having color blue requirement for YT and fix EoYT triggering on both player's turn.


x89rt2:
Also fix EoYT unsuspending itself

hooperk (approved):

PR #1132, opened 2026-01-07
Update BT24_023.cs

x89rt2:
Attempt to fix ByEffect

### DIFF
--- a/CardEffect/BT24/Blue/BT24_023.cs
+++ b/CardEffect/BT24/Blue/BT24_023.cs
@@ -78,9 +78,16 @@ namespace DCGO.CardEffects.BT24
                     || permanent.IsTamer);
             }
 
+            int ByEffect = 0;
+
             bool CanActivateConditionShared(Hashtable hashtable)
             {
                 return CardEffectCommons.IsExistOnBattleAreaDigimon(card);
+
+                if (CardEffectCommons.IsByEffect(hashtable, null))
+                {
+                    ByEffect = 1;
+                }
             }
 
             IEnumerator ActivateCoroutineShared(Hashtable hashtable, ActivateClass activateClass)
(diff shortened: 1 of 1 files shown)


CardEffect/BT24/Blue/BT24_023.cs
```diff
@@ -78,9 +78,16 @@ bool CanSelectPermanentCondition1(Permanent permanent)
...
+            int ByEffect = 0;
+
             bool CanActivateConditionShared(Hashtable hashtable)
             {
                 return CardEffectCommons.IsExistOnBattleAreaDigimon(card);
+
+                if (CardEffectCommons.IsByEffect(hashtable, null))
+                {
+                    ByEffect = 1;
+                }
```
hooperk, line 88: This needs to go before the line starting with "return", as return exits the method. Also, use bool instead of int, setting it equal to false initially and then set to true instead of setting to 1

CardEffect/BT24/Blue/BT24_023.cs
```diff
@@ -105,7 +112,7 @@ IEnumerator ActivateCoroutineShared(Hashtable hashtable, ActivateClass activateC
 
                 yield return ContinuousController.instance.StartCoroutine(selectPermanentEffect.Activate());
 
-                if (CardEffectCommons.IsByEffect(hashtable, null))
+                if (int ByEffect = 1)
```
hooperk: This is actually defining a new variable int and setting it to 1.
The declaration of type (int) should only appear the first time you reference it.
And you need to use double equals "==" for comparison. Single equals is assignment, meaning "set to this new value".


However, if this was a boolean instead you just write "if(ByEffect)"

hooperk (changes requested):

hooperk (approved):

PR #1137, opened 2026-01-07
Update BT23_029.cs

x89rt2:
Fix weird trigger timings.

### DIFF
--- a/CardEffect/BT23/Yellow/BT23_029.cs
+++ b/CardEffect/BT23/Yellow/BT23_029.cs
@@ -70,7 +70,7 @@ namespace DCGO.CardEffects.BT23
 
                 bool PermanentCondition(Permanent permanent)
                 {
-                    return CardEffectCommons.IsPermanentExistsOnOwnerBattleAreaDigimon(permanent, card)
+                    return CardEffectCommons.IsPermanentExistsOnOwnerBattleArea(permanent, card)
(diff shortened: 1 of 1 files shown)


CardEffect/BT23/Yellow/BT23_029.cs
```diff
@@ -70,7 +70,7 @@ bool CanActivateCondition(Hashtable hashtable)
 
                 bool PermanentCondition(Permanent permanent)
                 {
-                    return CardEffectCommons.IsPermanentExistsOnOwnerBattleAreaDigimon(permanent, card)
+                    return CardEffectCommons.IsPermanentExistsOnOwnerBattleArea(permanent, card)
                         && (permanent.TopCard.EqualsTraits("Beast") || permanent.TopCard.EqualsTraits("Beastkin") || permanent.TopCard.HasCSTraits);
```
hooperk, line 75: I think we'll need an (permanent.IsTamer || permanent.IsDigimon) since I think Delay Options enter the battle area too

x89rt2: hasplaycost?

hooperk: Hmm. Yeah that should actually work as a shorthand since these cards are any one that can be played

hooperk: Actually no, do not do that as then it would not correctly trigger for qualifying tokens. Just do (permanent.IsTamer || permanent.IsDigimon)

hooperk (approved):

PR #1140, opened 2026-01-08
Update BT24_059.cs

x89rt2:
Fix ESS by using working BT24-074 SkullSeadramon

### DIFF
--- a/CardEffect/BT24/Black/BT24_059.cs
+++ b/CardEffect/BT24/Black/BT24_059.cs
@@ -250,13 +250,11 @@ namespace DCGO.CardEffects.BT24
                         "Select 1 card to place on bottom of digivolution cards.",
                         "The opponent is selecting 1 card to place on bottom of digivolution cards.");
 
-                    yield return ContinuousController.instance.StartCoroutine(selectPermanentEffect.Activate());
-                    CardSource placedCard = null;
-                    
+                     yield return ContinuousController.instance.StartCoroutine(selectPermanentEffect.Activate());
+
                     IEnumerator SelectPermanentCoroutine(Permanent permanent)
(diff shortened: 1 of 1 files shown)


CardEffect/BT24/Black/BT24_059.cs
```diff
@@ -250,21 +250,19 @@ IEnumerator ActivateCoroutine(Hashtable hashtable)
...
-                        selectedCards.Add(placedCard);
+                        selectedCards.Add(permanent.TopCard);
 
                         yield return ContinuousController.instance.StartCoroutine(new IPlacePermanentToDigivolutionCards(
                             new List<Permanent[]>() { new Permanent[] { permanent, card.PermanentOfThisCard() } },
                             false,
                             activateClass).PlacePermanentToDigivolutionCards());
                     }
 
-                    if (card.PermanentOfThisCard().DigivolutionCards.Contains(placedCard))
```
hooperk, line 267: This is the tried and tested way to confirm the placing was not prevented, and this change would mean it counted as successful even if a prevention effect stopped it leaving to become a digivolution card. What bug is this looking to fix?

x89rt2: Previous code was not actually unsuspending. With that said, I also need to get skullsea interaction with tucking a token checked so this might need more changes.

hooperk: Looking at older cards like BT15-029 MegaSeadramon, I'm ok to let this in and try to fix the unsuspend issue, and we can look into this issue of whether or not it and similar cards deals with the protection scenario later

hooperk (approved):

PR #1142, opened 2026-01-08
Throne room cards

hooperk:
2 commits:
First reduces unnecessary conditions to try to cause less lag
Second ensures the Added alliance effect is correctly marked as be an effect added by this card

### DIFF
--- a/CardEffect/BT24/Blue/BT24_090.cs
+++ b/CardEffect/BT24/Blue/BT24_090.cs
@@ -84,12 +84,7 @@ namespace DCGO.CardEffects.BT24
                 {
                     if (effectTiming == EffectTiming.OnAllyAttack)
                     {
-                        bool Condition()
-                        {
-                            return CardSourceCondition(cardSource);
-                        }
-
-                        effects.Add(CardEffectFactory.AllianceSelfEffect(false, cardSource, Condition));
+                        effects.Add(CardEffectFactory.AllianceSelfEffect(false, cardSource, null));
                     }
 
                     return effects;
--- a/CardEffect/BT24/Green/BT24_094.cs
+++ b/CardEffect/BT24/Green/BT24_094.cs
@@ -48,8 +48,7 @@ namespace DCGO.CardEffects.BT24
                 #region +2k DP
                 bool CanUseCondition()
                 {
-                    return CardEffectCommons.IsExistInSecurity(card, false) &&
-                           CardEffectCommons.HasMatchConditionPermanent(PermanentCondition);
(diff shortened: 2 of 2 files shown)


CardEffect/BT24/Blue/BT24_090.cs
```diff
@@ -84,12 +84,7 @@ List<ICardEffect> GetEffects1(CardSource cardSource, List<ICardEffect> effects,
                 {
                     if (effectTiming == EffectTiming.OnAllyAttack)
                     {
-                        bool Condition()
-                        {
-                            return CardSourceCondition(cardSource);
-                        }
-
-                        effects.Add(CardEffectFactory.AllianceSelfEffect(false, cardSource, Condition));
+                        effects.Add(CardEffectFactory.AllianceEffect(cardSource.PermanentOfThisCard(), false, Condition, addSkillClass1, null));
```
Michael8818: This whole effect isn't correct. copy what EX9-033 does while changing conditions, this goes for both cards.

Michael8818 (changes requested):

CardEffect/BT24/Blue/BT24_090.cs
```diff
@@ -84,12 +84,7 @@ List<ICardEffect> GetEffects1(CardSource cardSource, List<ICardEffect> effects,
                 {
                     if (effectTiming == EffectTiming.OnAllyAttack)
                     {
-                        bool Condition()
-                        {
-                            return CardSourceCondition(cardSource);
-                        }
-
-                        effects.Add(CardEffectFactory.AllianceSelfEffect(false, cardSource, Condition));
+                        effects.Add(CardEffectFactory.AllianceSelfEffect(false, cardSource, null));
```
hooperk, line 87: For both: AddSkillEffect already checked CardSourceCondition for every card, so no need to recheck it

CardEffect/BT24/Green/BT24_094.cs
```diff
@@ -48,8 +48,7 @@ bool PermanentCondition(Permanent permanent)
                 #region +2k DP
                 bool CanUseCondition()
                 {
-                    return CardEffectCommons.IsExistInSecurity(card, false) &&
-                           CardEffectCommons.HasMatchConditionPermanent(PermanentCondition);
+                    return CardEffectCommons.IsExistInSecurity(card, false);
```
hooperk, line 51: No need to check we have a permanent that matches the condition and then recheck again as part of the static effect

Michael8818 (approved):

PR #1146, opened 2026-01-08
Fix BT23-084 Not considered to have an inheritable effect

hooperk:
Erika is not showing as having an Inheritable effect for BT7-040 Neemon. Trying to add a dummy inherited effect


x89rt2:
if works, reminder to fix bt7 JP

Michael8818 (dismissed):
i mean it should work do me a favor and also add activateClass.SetAsBackgroundEffect();

x89rt2 (approved):

Michael8818 (approved):

PR #1147, opened 2026-01-09
Update BT24_078.cs

x89rt2:
Playing ludicrous amounts of play cost. Used BT14-068 Brigadramon as a base to possibly fix.


CardEffect/BT24/Purple/BT24_078.cs
```diff
@@ -107,6 +107,8 @@ bool CanSelectCardCondition(CardSource cardSource)
                 {
                     return (cardSource.EqualsTraits("Evil")
                             || cardSource.EqualsTraits("Fallen Angel"))
+                        && cardSource.GetCostItself <= maxCost
```
hooperk, line 110: maxCost is not defined within this method

hooperk (changes requested):

PR #1149, opened 2026-01-09
Fix BT24-078 Can select too many cards

hooperk:
Copied more from EX8_064

### DIFF
--- a/CardEffect/BT24/Purple/BT24_078.cs
+++ b/CardEffect/BT24/Purple/BT24_078.cs
@@ -117,18 +117,19 @@ namespace DCGO.CardEffects.BT24
 
                     List<CardSource> selectedCards = new List<CardSource>();
                     int totalCost = 4 * (1 + (card.Owner.Enemy.TrashCards.Count / 10));
+                    int maxCount = CardEffectCommons.MatchConditionOwnersCardCountInTrash(CanSelectCardCondition);
 
                     SelectCardEffect selectCardEffect = GManager.instance.GetComponent<SelectCardEffect>();
 
                     selectCardEffect.SetUp(
                         canTargetCondition: CanSelectCardCondition,
                         canTargetCondition_ByPreSelecetedList: CanTargetCondition_ByPreSelecetedList,
-                        canEndSelectCondition: null,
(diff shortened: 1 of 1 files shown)


CardEffect/BT24/Purple/BT24_078.cs
```diff
@@ -117,18 +117,19 @@ IEnumerator ActivateCoroutine(Hashtable hashtable)
 
                     List<CardSource> selectedCards = new List<CardSource>();
                     int totalCost = 4 * (1 + (card.Owner.Enemy.TrashCards.Count / 10));
+                    int maxCost = CardEffectCommons.MatchConditionOwnersCardCountInTrash(CanSelectCardCondition);
```
x89rt2: First instinct is this doesn't seem right? Does this check each individual card's play cost?

Michael8818: that function will return the number of cards in trash that meet the condition passed in

x89rt2: That's what I thought. It's morning, but that doesn't seem to match what the equations at the bottom are trying to do. I could be wrong though.

hooperk: Oh typo, this should be maxCount, which is just the maximum cards this could possibly play. It needs to be given some upper bound, the conditions further below will restrict that to a maximum cost

x89rt2: hm.... I'm not sure if this works? what if we have 3 play cost worth of plays left, wouldn't current code allow you to select a 4 cost or higher? I feel like the canselectcardcondition needs an equation in it

hooperk: It would, and then you would not be able to accept that choice due to the CanEndSelectCardCondition added below

hooperk: In theory the preselect condition should stop it being a choice but it does not appear to

x89rt2: I see, so the end selection button would disappear. Alright, lets try it.

x89rt2 (approved):

PR #1153, opened 2026-01-09
Update BT15_054.cs

### LINKED ISSUE #243: BT15-054 Rosemon (X Antibody)
Her second Opponent's Turn Triggers and Activates when Opponent's digimon moves out of Breeding Area although to no effect if there is no Rosemon or X Antibody under her. Would be better if it just didn't trigger if she doesn't have said cards under her.
https://discord.com/channels/1095717982522585098/1229224709007343687/1407088257967198228

x89rt2:
Move source check from activatecoroutine to canactivatecondition

### DIFF
--- a/CardEffect/BT15/Green/BT15_054.cs
+++ b/CardEffect/BT15/Green/BT15_054.cs
@@ -253,7 +253,10 @@ namespace DCGO.CardEffects.BT15
                 {
                     if (CardEffectCommons.IsExistOnBattleArea(card))
                     {
-                        return true;
+                        if (card.PermanentOfThisCard().DigivolutionCards.Count((cardSource) => cardSource.CardNames.Contains("Rosemon") || cardSource.CardNames.Contains("X Antibody") || cardSource.CardNames.Contains("XAntibody")) >= 1)
+                        {
+                            return true;
+                        }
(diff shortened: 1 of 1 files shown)


CardEffect/BT15/Green/BT15_054.cs
```diff
@@ -253,7 +253,10 @@ bool CanActivateCondition(Hashtable hashtable)
                 {
                     if (CardEffectCommons.IsExistOnBattleArea(card))
                     {
-                        return true;
+                        if (card.PermanentOfThisCard().DigivolutionCards.Count((cardSource) => cardSource.CardNames.Contains("Rosemon") || cardSource.CardNames.Contains("X Antibody") || cardSource.CardNames.Contains("XAntibody")) >= 1)
```
hooperk: Use EqualsCardName for all 3 of these

hooperk: As a side note: doing so actually removes the need to separately check the last condition as checking without spaces is one of the things that method does more than just CardNames.Contains

hooperk (changes requested):

hooperk (approved):

PR #1161, opened 2026-01-09
ST21-09 Unregister OPT use if no processing occured

hooperk:


x89rt2:
Does this handle twice per turns when those are fixed?

hooperk:
> Does this handle twice per turns when those are fixed?

Yes. Twice per turn would be based on the Card Effect being in this particular list 2 times already, where currently Once Per turns only check if it appears once. Remove only removes exactly one instance of it from the list (the first it finds), so even with twice per turn this logic will work.

PR #1164, opened 2026-01-09
Update EX6_072.cs

### LINKED ISSUE #332: EX6-072 Mega Digimon Assembly! - Bug #1
Does not let you DNA into Milleni with a Kimera in hand
https://discord.com/channels/839041812932329512/1408945785902469120

### LINKED ISSUE #430: EX6-072 Mega Digimon Assembly! - Bug #3
Stops you from digivolving ever again if you attempt to use while your only legal targets are digivolve locked via MegaGargo Ace.
https://discord.com/channels/839041812932329512/1239959030819196968/1419821749419245748

x89rt2:
Attempt to fix for Milleni DNA of Kimera in hand

### DIFF
--- a/CardEffect/EX6/White/EX6_072.cs
+++ b/CardEffect/EX6/White/EX6_072.cs
@@ -72,8 +72,7 @@ namespace DCGO.CardEffects.EX6
 
                 bool HasLevel6HandSource(CardSource cardSource)
                 {
-                    return cardSource.IsDigimon &&
-                           cardSource.HasLevel && cardSource.Level == 6;
+                    return true;
                 }
 
                 bool HasLevel6Permanent(Permanent permanent)
--- a/CardEffect/EX6/White/EX6_072.cs
+++ b/CardEffect/EX6/White/EX6_072.cs
@@ -81,7 +81,10 @@ namespace DCGO.CardEffects.EX6
                     {
                         if (permanent.TopCard.HasLevel && permanent.Level == 6)
                         {
-                            return true;
+                            if (permanent.CanPlayCardTargetFrame(card.PermanentOfThisCard().PermanentFrame, true, activateClass);
+                            {
+                                return true;
+                            }
                         }
                     }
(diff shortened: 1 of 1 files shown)


x89rt2:
Attempt to fix megagargo ace digivolution bug, check line 84 and tell me correct code.

CardEffect/EX6/White/EX6_072.cs
```diff
@@ -82,7 +81,10 @@ bool HasLevel6Permanent(Permanent permanent)
                     {
                         if (permanent.TopCard.HasLevel && permanent.Level == 6)
                         {
-                            return true;
+                            if (permanent.CanPlayCardTargetFrame(card.PermanentOfThisCard().PermanentFrame, true, activateClass);
```
hooperk: CanPlayCardTargetFrame is a method of CardSource, not permanent, and currently this is asking could you digivolve the permanent into itself. I don't think you should include this check, as you need to have the target of the digivolution already chosen for any such check

hooperk (changes requested):

CardEffect/EX6/White/EX6_072.cs
x89rt2: Well, we do have a bug where the game soft locks if you try to dna with a megagargo blocked guy. We need something somewhere.

hooperk: try then `!card.CanNotEvolve(permanent)`
It'll check for can not evolve effects like megagargo

hooperk: It is part of what CanPlayCardTargetFrame checks, but is the only relevant part as the rest of that method is checking for if the card is a valid digivolution target, if you can pay the cost, etc.

hooperk: Sorry there was a typo in that suggestion, fixed now

hooperk (approved):

PR #1166, opened 2026-01-09
Fix BT18-015 & BT18-073 preferring top left corner

hooperk:


x89rt2:
I'm getting a Cannot convert method group 'PreferredFrame' to non-delegate type 'int'. Did you intend to invoke the method? on the                         int frameID = selectedCardSource.PreferredFrame;
line

x89rt2 (changes requested):
I'm getting a "Cannot implicitly convert type 'FieldCardFrame' to 'int'" on same line.

x89rt2 (approved):
Good enough for me.

PR #1167, opened 2026-01-09
Update BT5_109.cs

x89rt2:
First rough on getting Mega Digimon Fusion! working. I'm not sure what to put on line 50. Should be all that's pending.

### DIFF
--- a/CardEffect/BT5/White/BT5_109.cs
+++ b/CardEffect/BT5/White/BT5_109.cs
@@ -133,7 +133,7 @@ public class BT5_109 : CEntity_Effect
                     bool CardSourceCondition(CardSource cardSource)
                     {
                         return cardSource.HasLevel
-                            && cardSource.Level == 6;
+                            && cardSource.Level == 7;
                     }
 
                     bool RootCondition(SelectCardEffect.Root root)
--- a/CardEffect/BT5/White/BT5_109.cs
+++ b/CardEffect/BT5/White/BT5_109.cs
@@ -50,7 +50,7 @@ public class BT5_109 : CEntity_Effect
                 Permanent playedPermanent = ;
 
                 ActivateClass activateClass3 = new ActivateClass();
-                activateClass3.SetUpICardEffect("Delete the Digimon", CanUseCondition2, playedPermanent.TopCard);
+                activateClass3.SetUpICardEffect("Bottom deck the Digimon", CanUseCondition2, playedPermanent.TopCard);
                 activateClass3.SetUpActivateClass(CanActivateCondition1, ActivateCoroutine3, -1, false, EffectDiscription2());
                 activateClass3.SetEffectSourcePermanent(playedPermanent);
                 playedPermanent.UntilOwnerTurnEndEffects.Add(GetCardEffect);
--- a/CardEffect/BT5/White/BT5_109.cs
+++ b/CardEffect/BT5/White/BT5_109.cs
@@ -91,12 +91,12 @@ public class BT5_109 : CEntity_Effect
                     ContinuousController.instance.PlaySE(GManager.instance.GetComponent<Effects>().BuffSE);
 
                     ChangeCostClass changeCostClass = new ChangeCostClass();
-                    changeCostClass.SetUpICardEffect("Digivolution Cost -6", CanUseCondition1, card);
+                    changeCostClass.SetUpICardEffect("Digivolution Cost -6", CanUseCondition3, card);
                     changeCostClass.SetUpChangeCostClass(changeCostFunc: ChangeCost, cardSourceCondition: CardSourceCondition, rootCondition: RootCondition, isUpDown: isUpDown, isCheckAvailability: () => false, isChangePayingCost: () => true); card.Owner.UntilCalculateFixedCostEffect.Add((_timing) => changeCostClass);
 
                     yield return ContinuousController.instance.StartCoroutine(CardEffectCommons.ShowReducedCost(_hashtable1));
 
-                    bool CanUseCondition1(Hashtable hashtable)
+                    bool CanUseCondition3(Hashtable hashtable)
                     {
                         return true;
                     }
--- a/CardEffect/BT5/White/BT5_109.cs
+++ b/CardEffect/BT5/White/BT5_109.cs
@@ -30,9 +30,6 @@ public class BT5_109 : CEntity_Effect
 
             IEnumerator ActivateCoroutine(Hashtable _hashtable)
             {
-
-                #region Reduce evo cost
-
                 ContinuousController.instance.PlaySE(GManager.instance.GetComponent<Effects>().BuffSE);
 
                 yield return new WaitForSeconds(0.2f);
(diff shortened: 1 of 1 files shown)


CardEffect/BT5/White/BT5_109.cs
```diff
@@ -30,11 +30,224 @@ bool CanUseCondition(Hashtable hashtable)
...
+                {
+                    ContinuousController.instance.PlaySE(GManager.instance.GetComponent<Effects>().BuffSE);
+
+                    ChangeCostClass changeCostClass = new ChangeCostClass();
+                    changeCostClass.SetUpICardEffect("Digivolution Cost -6", CanUseCondition1, card);
+                    changeCostClass.SetUpChangeCostClass(changeCostFunc: ChangeCost, cardSourceCondition: CardSourceCondition, rootCondition: RootCondition, isUpDown: isUpDown, isCheckAvailability: () => false, isChangePayingCost: () => true); card.Owner.UntilCalculateFixedCostEffect.Add((_timing) => changeCostClass);
+
+                    yield return ContinuousController.instance.StartCoroutine(CardEffectCommons.ShowReducedCost(_hashtable1));
+
+                    bool CanUseCondition1(Hashtable hashtable)
```
hooperk: Line 71 is ALSO named the same and is still in scope for here, that is why the red line

x89rt2: That's the thing, I'm pretty sure HPU has the same thing, so I thought it might need to be the same here.

x89rt2: hm... I might be wrong actually. Did I confuse myself copying code somehow?

CardEffect/BT5/White/BT5_109.cs
```diff
@@ -30,11 +30,224 @@ bool CanUseCondition(Hashtable hashtable)
...
+                Func<EffectTiming, ICardEffect> getCardEffect1 = GetCardEffect1;
+                activateClass2.SetUpICardEffect("Remove Effect", CanUseCondition1, card);
+                activateClass2.SetUpActivateClass(null, ActivateCoroutine2, -1, false, "");
+                activateClass2.SetIsBackgroundProcess(true);
+                CardEffectCommons.AddEffectToPlayer(effectDuration: EffectDuration.UntilEachTurnEnd, card: card, cardEffect: null, timing: EffectTiming.None, getCardEffect: getCardEffect1);
+
+                Permanent playedPermanent = ;
+
+                ActivateClass activateClass3 = new ActivateClass();
+                activateClass3.SetUpICardEffect("Delete the Digimon", CanUseCondition2, playedPermanent.TopCard);
```
hooperk: Deck Bottom the digimon

CardEffect/BT5/White/BT5_109.cs
```diff
@@ -30,11 +30,224 @@ bool CanUseCondition(Hashtable hashtable)
...
+                        }
+
+                        return false;
+                    }
+
+                    bool CardSourceCondition(CardSource cardSource)
+                    {
+                        return cardSource.HasLevel
+                            && cardSource.Level == 6;
+                    }
```
hooperk, line 137: I believe CardCondition applies to the card being played or digivolved into, so it would be level 7. Also .IsLevel7 can replace these 2 conditions

x89rt2: I'm not sure we have an IsLevel7 programmed last time I checked.

hooperk: Oh then this method but == 7 will be fine

CardEffect/BT5/White/BT5_109.cs
```diff
@@ -30,11 +30,224 @@ bool CanUseCondition(Hashtable hashtable)
...
+                        }
+
+                        return Cost;
+                    }
+
+                    bool PermanentsCondition(List<Permanent> targetPermanents)
+                    {
+                        if (targetPermanents != null)
+                        {
+                            if (targetPermanents.Count(PermanentCondition) >= 1)
```
hooperk, line 124: I think this should be ==1 in order to correctly be "when 1 of your level 6 digimon..."

CardEffect/BT5/White/BT5_109.cs
x89rt2: Ok, all that's pending is what to put in line 50. It's been a couple of days so I don't remember which of the 2 reference cards was that had it but 1 was Hidden Potential Unleashed, don't remember the second right now.

CardEffect/BT5/White/BT5_109.cs
```diff
@@ -30,11 +30,224 @@ bool CanUseCondition(Hashtable hashtable)
...
+
+                    bool isUpDown()
+                    {
+                        return true;
+                    }
+                }
+
+                ICardEffect GetCardEffect(EffectTiming _timing)
+                {
+                    if (_timing == EffectTiming.BeforePayCost)
```
hooperk, line 152: If this effect goes off here, it will remove the cost reduction just before you would use int. Instead maybe this could go to when digivolving timing and remove the effect just after digivolution? EffectTiming.OnEnterFieldAnyone

CardEffect/BT5/White/BT5_109.cs
```diff
@@ -30,11 +30,224 @@ bool CanUseCondition(Hashtable hashtable)
...
+                        return activateClass1;
+                    }
+
+                    return null;
+                }
+
+
+                IEnumerator ActivateCoroutine2(Hashtable _hashtable1)
+                {
+                    if (CardEffectCommons.CanTriggerWhenPermanentWouldDigivolve(
```
hooperk, line 163: If this does move to when digivolving timing, this will need to change with it

CardEffect/BT5/White/BT5_109.cs
```diff
@@ -30,11 +30,224 @@ bool CanUseCondition(Hashtable hashtable)
...
+                        hashtable: _hashtable1,
+                        permanentCondition: PermanentCondition,
+                        cardCondition: null))
+                    {
+                        card.Owner.UntilEachTurnEndEffects.Remove(getCardEffect);
+                        card.Owner.UntilEachTurnEndEffects.Remove(getCardEffect1);
```
hooperk, line 169: You should probably also add the deck bottom effect as part of this effect.

CardEffect/BT5/White/BT5_109.cs
x89rt2: Doesn't matter cause you can only do 1 digivolution at a time. Also, this "one" might be an "any".

CardEffect/BT5/White/BT5_109.cs
x89rt2: whoops

CardEffect/BT5/White/BT5_109.cs
hooperk: Jogress would count and is 2 permanents. Although currently there isn't a jogress with a non-0 cost for this to matter for, it does mean that jogressed digimon would be given the deck bounce effect despite not meeting this card's condition

x89rt2: Am I still missing something or isn't that fine? since one means any? For the record, my brain is not really on right now. https://discord.com/channels/681578268729540663/749146995708395601/1461739750250647623

PR #1176, opened 2026-01-10
Implement EX11-070 Unchained with CEF method for playing Mind Link Tamer and CEC method for starting a DNA Digivolution via effect

hooperk:


CardEffect/EX11/White/EX11_070.cs
```diff
@@ -0,0 +1,257 @@
...
+                    ChangeDPClass changeDPClass = new ChangeDPClass();
+                    changeDPClass.SetUpICardEffect("Can't have less than 1000 DP", CanUseCondition, card);
+                    changeDPClass.SetUpChangeDPClass(ChangeDP: ChangeDP, permanentCondition: PermanentCondition, isUpDown: false, isMinusDP: () => false);
+                    changeDPClass.SetIsInheritedEffect(true);
+                    cardEffects.Add(changeDPClass);
+
+                    bool CanUseCondition(Hashtable hashtable)
+                    {
+                        CardEffectCommons.IsExistOnBattleAreaDigimon(card);
+                    }
```
x89rt2: hm.... wouldn't it be more efficient to have the maquina in text on the condition so it doesn't process anything after this? same for can't trash

x89rt2: also, do these need to be different canuseconditions? can't you share to save processing power?

hooperk: You're right on the first, and it is part of the trigger condition either way.
It won't actually save processing power, it needs to be checked for each effect either way, but it will be nicer looking code

CardEffect/EX11/White/EX11_070.cs
```diff
@@ -0,0 +1,255 @@
...
+                    #endregion
+                    #region Mind Link
+                    yield return ContinuousController.instance.StartCoroutine(
+                        new MindLinkClass(
+                            tamer: card.PermanentOfThisCard(),
+                            digimonCondition: IsMyProperDigimon,
+                            activateClass: activateClass
+                        ).MindLink()
+                    );
+                    #endregion
```
x89rt2, line 84: is this in the right location?

hooperk: Yes, The Mind Link is the second part of the End of your Turn, and happens regardless of if you DNA or not

x89rt2: Does this have an option to choose which digimon to mind link to and to say no?

hooperk: Yes, it has the full selection dialog with canSelectNo

Michael8818:
too big of a merge conflict since I didn't write it I don't see where i need to merge the lines. Your going to need to resolve this conflict first.

hooperk:
Going to recreate this from internal repo so I can better check it compiles and everything

PR #1179, opened 2026-01-10
Fix Game crash from Throne Rooms

hooperk:


hooperk:
IsExistOnBattleAreaDigimon should have already checked card.PermanentOfThisCard() is not null, so this is not the issue

PR #1185, opened 2026-01-11
Update BT24_029.cs

x89rt2:
possibly fix still playing egg?

### DIFF
--- a/CardEffect/BT24/Blue/BT24_029.cs
+++ b/CardEffect/BT24/Blue/BT24_029.cs
@@ -219,8 +219,7 @@ namespace DCGO.CardEffects.BT24
 
                 bool CardCondition(CardSource cardSource)
                 {
-                    return cardSource.HasPlayCost && cardSource.BasePlayCostFromEntity <= 5
-                        && cardSource.HasLevel
+                    return cardSource.HasPlayCost && cardSource.GetCostItself <= 5 && cardSource.GetCostItself >=1
                         && cardSource.HasTSTraits
                         && CardEffectCommons.CanPlayAsNewPermanent(cardSource, false, activateClass, root: SelectCardEffect.Root.DigivolutionCards);
                 }
--- a/CardEffect/BT24/Blue/BT24_029.cs
+++ b/CardEffect/BT24/Blue/BT24_029.cs
@@ -33,7 +33,7 @@ namespace DCGO.CardEffects.BT24
             {
                 bool CardCondition(CardSource cardSource)
                 {
-                    return cardSource.HasPlayCost && cardSource.BasePlayCostFromEntity <= 5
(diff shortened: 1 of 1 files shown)


CardEffect/BT24/Blue/BT24_029.cs
```diff
@@ -220,6 +220,7 @@ bool CanActivateCondition(Hashtable hashtable)
                 bool CardCondition(CardSource cardSource)
                 {
                     return cardSource.HasPlayCost && cardSource.BasePlayCostFromEntity <= 5
+                        && cardSource.HasLevel
```
hooperk: Digitama have a level. HasPlayCost also already checks if it's a digiegg. This is actually a card data issue with the level 2s. However, while you're looking at this card could you get it to use cardSource.GetCostItself <= 5? that's the recommended way to do that check now

CardEffect/BT24/Blue/BT24_029.cs
```diff
@@ -219,7 +219,7 @@ bool CanActivateCondition(Hashtable hashtable)
 
                 bool CardCondition(CardSource cardSource)
                 {
-                    return cardSource.HasPlayCost && cardSource.BasePlayCostFromEntity <= 5
+                    return cardSource.HasPlayCost && cardSource.GetCostItself <= 5 && cardSource.GetCostItself >=1
```
hooperk: rather than cardSource.GetCostItself >=1, `!cardSource.IsDigiEgg`

hooperk (changes requested):

CardEffect/BT24/Blue/BT24_029.cs
x89rt2: I don't like that. If there's ever a digiegg that has a play cost, that would stop it, no?

hooperk: True. But right now if we ever get a 0 cost digimon (an ace rookie?) We would have the same problem with it

hooperk: We can also just do neither, sounds like Michael is gonna update the egg data

x89rt2:
Changed with the assumption all appropriate eggs play cost card data gets changed from 0 to -1.

CardEffect/BT24/Blue/BT24_029.cs
```diff
@@ -219,7 +219,7 @@ bool CanActivateCondition(Hashtable hashtable)
 
                 bool CardCondition(CardSource cardSource)
                 {
-                    return cardSource.HasPlayCost && cardSource.BasePlayCostFromEntity <= 5
+                    return cardSource.HasPlayCost && cardSource.GetCostItself <= 5
```
hooperk, line 222: Could you also fix the cardCondition of the first effect the same way?

hooperk (approved):

PR #1187, opened 2026-01-11
Update BT24_018.cs

x89rt2:
Copied from BT14-075 Devimon OD. Current issue is it properly doesn't reveal sec, but it can't select face downs. Also, there's a chance this one doesn't let it select face ups?

### DIFF
--- a/CardEffect/BT24/Red/BT24_018.cs
+++ b/CardEffect/BT24/Red/BT24_018.cs
@@ -7,7 +7,7 @@ using UnityEngine;
 // Styracomon
 namespace DCGO.CardEffects.BT24
 {
-    public class BT24_018: CEntity_Effect
+    public class BT24_018 : CEntity_Effect
     {
         public override List<ICardEffect> CardEffects(EffectTiming timing, CardSource card)
         {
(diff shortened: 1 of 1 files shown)


CardEffect/BT24/Red/BT24_018.cs
```diff
@@ -96,14 +96,23 @@ bool CanActivateCondition(Hashtable hashtable)
...
                 {
                     #region Trash Security
+
                     if (card.Owner.Enemy.SecurityCards.Count >= 1)
                     {
+                        if (card.Owner.isYou)
+                        {
+                            foreach (CardSource cardSource in card.Owner.Enemy.SecurityCards)
+                            {
+                                cardSource.SetReverse();
```
hooperk, line 106: All this is doing is setting all of the security to be face down, which means the user won't be able to see the face up cards and they'll now be face down after the effect which is not right

hooperk: I had to dive into SelectHandEffect to find it but I believe I have the fix https://github.com/DCGO2/DCGO-Card-Scripts/pull/1190

PR #1191, opened 2026-01-11
Reset used blitz if a new open state is reached in main phase

hooperk:
Currently UsedBlitz is only reset if something sets the memory to a specific value, including passing turn. Adding a check to reset it if the game returns to an open main phase state with memory >= 0.


x89rt2:
Probably best to wait for post bt24 to implement.

PR #1197, opened 2026-01-11
Fix BT24-013 not triggering OnDiscardHand

hooperk:


hooperk:
Can see from other cards that do work, this approach is not the solution

PR #1198, opened 2026-01-11
Beta fix: New way of performing Assembly

hooperk:


Michael8818 (approved):
looks fine but this could break assembly so test in detail

PR #1205, opened 2026-01-12
Fix BT18-034 Lucemon EoYT to be able to digivolve with tokens

THE0Patate:
Issue 91
The code checks if the card has been added to the security. However, tokens cannot be added to the security. So, only checking if the card has left the battle area should do the job.

### DIFF
--- a/CardEffect/BT18/Yellow/BT18_034.cs
+++ b/CardEffect/BT18/Yellow/BT18_034.cs
@@ -411,7 +411,8 @@ namespace DCGO.CardEffects.BT18
 
                                     if (selectedPermanent.TopCard == null)
                                     {
-                                        if (!CardEffectCommons.IsExistOnBattleArea(securityCard))
+                                        if (securityCard.Owner.SecurityCards.Contains(securityCard) 
+                                            || (securityCard.IsToken && !CardEffectCommons.IsExistOnBattleArea(securityCard)))
                                         {
                                             if (CardEffectCommons.HasMatchConditionOwnersCardInTrash(card, IsCardLucemonChaosModeCondition))
                                             {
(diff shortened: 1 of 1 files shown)


x89rt2:
Hm... I'm not sure. With token protection or a replacement effect like queenbee or mother eater, this method could become an issue in the future. Let me know if you think otherwise. We are having a similar issue with the blue cards that tuck to unsuspend that someone is working on that probably will work for this.

THE0Patate:
Oh true, I didn't think of any sort of protection, but I'm still not sure if that matter, since the Coroutine IPutSecurityPermanent.PutSecurity() happens before this check (line 407) and it triggers potentials protections effets. Mabye I'm wrong about this though. First time helping you guys with bugs. Anyways, I'm will let him check this.

CardEffect/BT18/Yellow/BT18_034.cs
```diff
@@ -411,7 +411,7 @@ IEnumerator SelectPermanentCoroutine(Permanent permanent)
 
                                     if (selectedPermanent.TopCard == null)
                                     {
-                                        if (securityCard.Owner.SecurityCards.Contains(securityCard))
+                                        if (!CardEffectCommons.IsExistOnBattleArea(securityCard))
```
hooperk: I think both arguments are correct, so lets do this alternative only for tokens. I actually believe this should be a standard way of checking, I may make a CardEffectCommons method for it later
```suggestion
                                        if (securityCard.Owner.SecurityCards.Contains(securityCard) 
                                            || (securityCard.IsToken && !CardEffectCommons.IsExistOnBattleArea(securityCard)))
```

Michael8818 (approved):

PR #1206, opened 2026-01-12
Beta Fix: Fix BT24-102: Can tell what kind of ability is being used

hooperk:


CardEffect/BT24/White/BT24_102.cs
```diff
@@ -179,8 +176,15 @@ IEnumerator SelectCardCoroutine(Permanent permanent)
...
                             else
                             {
-                                List<SkillInfo> skillInfos = candidateEffects
-                                    .Map(cardEffect => new SkillInfo(cardEffect, null, EffectTiming.None));
+                                List<SkillInfo> skillInfos = new List<SkillInfo>();
+                                
+                                foreach (ICardEffect effect in candidateEffects)
+                                {
+                                    ICardEffect cardEffect = new ChangeBaseDPClass();
+                                    cardEffect.SetUpICardEffect((effect.IsOnPlay ? "On Play: " : "When Digivolving: ") + effect.EffectName, null, effect.EffectSourceCard);
```
hooperk, line 184: I was considering if it would fit if I just put the EffectDiscription as the skill info text?

x89rt2: It's normally in like [   ] no?

hooperk: It would be with the Effect Description, but here I am using the card effect name and still trying to keep the character count low so it all fits

Michael8818 (approved):

PR #1211, opened 2026-01-14
Fix for ST14-11 Ai & Mako - Bug #1

### LINKED ISSUE #385: ST14-11 Ai & Mako - Bug #1
Triggers from tamers digivolving into digimon - should be similar to tsunomon and snowagumon?

Zenivik:
Fix for YT triggering from tamers digivolving into digimon.
Should now check only digimon in battle area instead of everything in battle area.


Michael8818 (approved):
lgtm

x89rt2:
issue #1211 385

PR #1228, opened 2026-01-15
Beta Fix: Strip non-rendered characters from input

hooperk:
Specifically looking to fix LM-012 Lamortmon, but in case it shows up on any other cards 
image


hooperk:
Not how it is done

PR #1231, opened 2026-01-15
Beta Fix: Fix allow link cards to check breeding area for legality & Ensure they only link to Digimon

hooperk:


Scripts/CardSource.cs
```diff
@@ -2946,16 +2946,34 @@ public bool CanLink(bool PayCost)
                     return false;
                 }
             }
-
-            if (Owner.GetBattleAreaDigimons().Count >= 1)
```
hooperk, line 2950: The TL;DR is that because of this line, CanLink was only ever checking when you had Battle Area Permanents, and would return false otherwise. Now it more intelligently works based on whether or not the breeding are is allowed

CardEffect/BT24/Blue/BT24_091.cs
```diff
@@ -86,8 +86,8 @@ bool CanUseCondition(Hashtable hashtable)
 
                 bool CanSelectPermanentCondition(Permanent permanent)
                 {
-                    return (CardEffectCommons.IsPermanentExistsOnOwnerBattleArea(permanent, card) || CardEffectCommons.IsPermanentExistsOnOwnerBreedingArea(permanent, card))
-                        &&  card.CanLinkToTargetPermanent(permanent, false);
+                    return CardEffectCommons.IsOwnerPermanent(permanent, card)
```
hooperk: All permanents are either in the breeding area or battle area, so simplified this down to just checking it belongs to the card Owner

CardEffect/BT24/Blue/BT24_091.cs
```diff
@@ -38,7 +38,7 @@ bool CardCondition(CardSource cardSource)
             {
                 static bool PermanentCondition(Permanent targetPermanent)
                 {
-                    return targetPermanent.TopCard.HasTSTraits;
+                    
                 }
```
Michael8818, line 42: looks like this is just empty, it either isn't used (so should be removed) or must return a bool value

Michael8818 (changes requested):

Michael8818 (approved):

PR #1255, opened 2026-01-16
Update BT24_040.cs

x89rt2:
Used other card examples to fix by not your effect.


x89rt2:
Do we want to revert the previous change?

       bool CanPlaceToSecurityCondition(Permanent permanent)
                {
                    foreach (Permanent removed in removedPermanents)
                    {
                        if (removed != permanent
                            && CardEffectCommons.IsPermanentExistsOnBattleAreaDigimon(permanent)
                            && permanent.DigivolutionCards.Count == 0)
                            return true;
                    }
                    return false;
               }

hooperk (approved):

PR #1256, opened 2026-01-16
Partial Fix: DigiXros/Assembly not working when multiple cards played

hooperk:
Currently if the first Digimon in the list of CardSources is not the one for whom the digixros/assembly is intended, it will not work at all as it gets reset. Now the reset has been moved until after all Digimon are played to ensure it works for 1 of the played Digimon. However, fixing to ensure it works for all if multiple with digixros were played at once would require a lot more work.

### DIFF
--- a/Scripts/CardController.cs
+++ b/Scripts/CardController.cs
@@ -1613,12 +1613,13 @@ public class PlayPermanentClass
                 }
             }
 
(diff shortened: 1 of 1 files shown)


x89rt2:
@hooperk This got purged due to it being merged with Thrones fix. Resubmit.

PR #1267, opened 2026-01-16
Update ST22_06.cs

x89rt2:
Add a little more insurance that removal works properly when involving tokens.


hooperk:
Thought this one was the original one this change was done for. If you identify any other let me know, I want to make a common method for this

hooperk (approved):

PR #1273, opened 2026-01-17
CEC method for checking a card was sent to security

hooperk:
Occurred to me Eggs also need to be considered

### DIFF
--- a/CardEffect/LM/Yellow/LM_020.cs
+++ b/CardEffect/LM/Yellow/LM_020.cs
@@ -63,8 +63,6 @@ namespace DCGO.CardEffects.LM
                 {
                     bool placedToSecurity = false;
 
-                    List<Permanent> selectedPermanents = new List<Permanent>();
-
                     SelectPermanentEffect selectPermanentEffect = GManager.instance.GetComponent<SelectPermanentEffect>();
 
                     selectPermanentEffect.SetUp(
@@ -88,21 +86,11 @@ namespace DCGO.CardEffects.LM
 
                     IEnumerator SelectPermanentCoroutine(Permanent permanent)
                     {
-                        selectedPermanents.Add(permanent);
-
-                        Player selectedOwner = permanent.TopCard.Owner;
-                        CardSource securityCard = permanent.TopCard;
-                        int beforeEffectSecurityCount = permanent.TopCard.Owner.SecurityCards.Count;
-
-                        yield return ContinuousController.instance.StartCoroutine(new IPutSecurityPermanent(permanent, CardEffectCommons.CardEffectHashtable(activateClass), toTop: true).PutSecurity());
-
-                        if (WasSentToSecurity(securityCard))
-                            placedToSecurity = true;
-
-                        yield return null;
+                        yield return ContinuousController.instance.StartCoroutine(
+                            CardEffectCommons.PlacePermanentInSecurityAndProcessAccordingToResult(permanent, activateClass, toTop: true, SuccessProcess));
                     }
 
-                    if (placedToSecurity)
+                    IEnumerator SuccessProcess(CardSource cardSource)
                     {
                         if (card.Owner.Enemy.SecurityCards.Count >= 1)
                         {
--- a/CardEffect/LM/Yellow/LM_020.cs
+++ b/CardEffect/LM/Yellow/LM_020.cs
@@ -90,16 +90,14 @@ namespace DCGO.CardEffects.LM
                     {
                         selectedPermanents.Add(permanent);
 
-                        if (!permanent.IsToken)
-                        {
-                            Player selectedOwner = permanent.TopCard.Owner;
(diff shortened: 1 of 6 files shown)


hooperk:
Also gives us a central place to make improvements if we think of a better way to check that accounts for protections

x89rt2:
There needs to be differentiation on if sent to own sec, sent to oppo sec, or sent to either sec.

hooperk:
> There needs to be differentiation on if sent to own sec, sent to oppo sec, or sent to either sec.

Cards can only ever be sent to their owners security. DCG philosophically does not do card stealing/lending

x89rt2:
I grants something like venusmon. It can send a card to sec from either player's field.

hooperk:
> I grants something like venusmon. It can send a card to sec from either player's field.

Yes, but always the owner of that card that is sent. This method checks if the card is in the owner of that card's security, it has no concept of the effect that placed it there or who owned that effect.

x89rt2:
@hooperk can you also add P-187 and LM-020?

CardEffect/LM/Yellow/LM_020.cs
```diff
@@ -90,16 +90,14 @@ IEnumerator SelectPermanentCoroutine(Permanent permanent)
                     {
                         selectedPermanents.Add(permanent);
 
-                        if (!permanent.IsToken)
-                        {
-                            Player selectedOwner = permanent.TopCard.Owner;
-                            int beforeEffectSecurityCount = permanent.TopCard.Owner.SecurityCards.Count;
+                        Player selectedOwner = permanent.TopCard.Owner;
+                        CardSource securityCard = permanent.TopCard;
+                        int beforeEffectSecurityCount = permanent.TopCard.Owner.SecurityCards.Count;
```
x89rt2: Lines 93 and 95 are not used.

Scripts/CardEffectCommons/CanUseEffects/WhendAddSecurity.cs
```diff
@@ -12,4 +12,29 @@ public static bool CanTriggerWhenAddSecurity(Hashtable hashtable, Func<Player, b
...
+        if (cardSource.IsToken)
+        {
+            if (!CardEffectCommons.IsExistOnBattleArea(topCard))
+            {
+                return true;
+            }
+        }
+        if (cardSource.IsDigiEgg)
+        {
+            if (!CardEffectCommons.IsExistOnBattleArea(topCard))
```
x89rt2: topCard on lines 24 and 31 are not defined, shouldn't they be cardSource? If they need to be defined on the cards themselves, we might want to do this in a slightly different way. I'm not 100% on the understanding yet so I'd need time to make any suggestions.

Scripts/CardEffectCommons/CanUseEffects/WhendAddSecurity.cs
```diff
@@ -12,4 +12,29 @@ public static bool CanTriggerWhenAddSecurity(Hashtable hashtable, Func<Player, b
         return CanTriggerWhenLoseSecurity(hashtable, playerCondition);
     }
     #endregion
+
```
x89rt2, line 15: Also, name of this file has a misspelling I think.

x89rt2 (changes requested):

Scripts/CardEffectCommons/CanUseEffects/WhendAddSecurity.cs
hooperk: It does but the file already exists and I don't want to mess with it

Scripts/CardEffectCommons/CanUseEffects/WhendAddSecurity.cs
x89rt2: Don't these still need a CardSource topCard = targetPermanent.TopCard;? or maybe make the one on line 46 a public bool?

hooperk: This method checks the card itself for its status. It is called by other methods, like the one below, which passes the card to check

x89rt2:
Primary look over looks good to me. I'll leave a final confirmation for Mike or Lewis.

Scripts/CardEffectCommons/CanUseEffects/WhendAddSecurity.cs
```diff
@@ -12,4 +12,58 @@ public static bool CanTriggerWhenAddSecurity(Hashtable hashtable, Func<Player, b
...
+        {
+            if (!CardEffectCommons.IsExistOnBattleArea(cardSource))
+            {
+                return true;
+            }
+        }
+
+        return false;
+    }
+    #endregion
```
x89rt2, line 39: So this is failing for some reason.

https://discord.com/channels/1095717982522585098/1392137159275708456/1463975556096987312

hooperk:
Closing in favour of https://github.com/DCGO2/DCGO-Card-Scripts/pull/1344

PR #1275, opened 2026-01-17
Update BT24_007.cs

x89rt2:
Not sure if it even needs trashedcards.count with this change.

Fix to plays card when it's no longer in trash.

### DIFF
--- a/CardEffect/BT24/Purple/BT24_007.cs
+++ b/CardEffect/BT24/Purple/BT24_007.cs
@@ -51,7 +51,7 @@ namespace DCGO.CardEffects.BT24
                 {
                     List<CardSource> trashedCards = CardEffectCommons.GetDiscardedCardsFromHashtable(hashtable).Filter(CardCondition);
 
-                    if (trashedCards.Count > 0 && CardEffectCommons.HasMatchConditionOwnersCardInTrash(card, trashedCards => trashCards.Contains(trashedCards)))
+                    if (CardEffectCommons.HasMatchConditionOwnersCardInTrash(card, trashCard=> trashedCards.Contains(trashCard)))
                     {
                         CardSource selectedCard = null;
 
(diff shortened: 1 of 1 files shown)


CardEffect/BT24/Purple/BT24_007.cs
```diff
@@ -51,7 +51,7 @@ IEnumerator ActivateCoroutine(Hashtable hashtable)
                 {
                     List<CardSource> trashedCards = CardEffectCommons.GetDiscardedCardsFromHashtable(hashtable).Filter(CardCondition);
 
-                    if (trashedCards.Count > 0)
+                    if (trashedCards.Count > 0 && CardEffectCommons.HasMatchConditionOwnersCardInTrash(card, CardCondition))
```
hooperk: replace CardCondition on this line with `trashCard => trashCards.Contains(trashCard)`
CardCondition would check for any Demon or Titan, this line will check instead if there is still a card that was one of the trashed cards that qualified

hooperk (changes requested):

CardEffect/BT24/Purple/BT24_007.cs
```diff
@@ -51,7 +51,7 @@ IEnumerator ActivateCoroutine(Hashtable hashtable)
                 {
                     List<CardSource> trashedCards = CardEffectCommons.GetDiscardedCardsFromHashtable(hashtable).Filter(CardCondition);
 
-                    if (trashedCards.Count > 0)
+                    if (trashedCards.Count > 0 && CardEffectCommons.HasMatchConditionOwnersCardInTrash(card, trashedCards => trashCards.Contains(trashedCards)))
```
hooperk:
```suggestion
                    if (CardEffectCommons.HasMatchConditionOwnersCardInTrash(card, trashCard=> trashedCards.Contains(trashCard)))
```
Typo with trashCard and trashedCards mixed up. Also we won't need to check the size of trashCards if we're checking the contents (.Contains on an empty list always returns false)

hooperk (changes requested):

hooperk (approved):

PR #1285, opened 2026-01-18
Create EX11_061.cs

### LINKED ISSUE #798: EX11-061 - Mirai Kinosaki
Image

x89rt2:


CardEffect/EX11/Yellow/EX11_061.cs
```diff
@@ -0,0 +1,191 @@
...
+{
+    public class EX11_061 : CEntity_Effect
+    {
+        public override List<ICardEffect> CardEffects(EffectTiming timing, CardSource card)
+        {
+            List<ICardEffect> cardEffects = new List<ICardEffect>();
+
+            #region Start of Turn
+
+            if (timing == EffectTiming.OnStartMainPhase)
```
hooperk, line 16: BTW, we have a convenience method so we don't have to write this each time:
```
if (timing == EffectTiming.OnStartMainPhase)
{
      cardEffects.Add(CardEffectFactory.Gain1MemoryTamerOpponentDigimonEffect(card));
}
```

CardEffect/EX11/Yellow/EX11_061.cs
```diff
@@ -0,0 +1,191 @@
...
+
+                string EffectDiscription()
+                {
+                    return "[Your Turn] When any of your Digimon digivolve into a [Puppet] trait Digimon, by suspending this Tamer, you may play 1 level 3 [Puppet] trait Digimon card from your hand without paying the cost. At turn end, delete the Digimon this effect played.";
+                }
+
+                bool CanUseCondition(Hashtable hashtable)
+                {
+                    return CardEffectCommons.CanTriggerWhenPermanentDigivolving(hashtable, permanent =>
+                    {
```
hooperk: This is a long condition, could you extract it to a bool PermanentCondition(Permanent permanent)?

CardEffect/EX11/Yellow/EX11_061.cs
```diff
@@ -0,0 +1,191 @@
...
+                            #region Delete Played Digimon
+
+                            Permanent selectedPermanent = selectCard.PermanentOfThisCard();
+
+                            ActivateClass activateClass1 = new ActivateClass();
+                            activateClass1.SetUpICardEffect("Delete this Digimon", CanUseCondition2, card);
+                            activateClass1.SetUpActivateClass(CanActivateCondition1, ActivateCoroutine1, -1, false, "");
+                            activateClass1.SetEffectSourcePermanent(selectedPermanent);
+                            CardEffectCommons.AddEffectToPlayer(effectDuration: EffectDuration.UntilEachTurnEnd, card: card, cardEffect: activateClass1, timing: EffectTiming.OnEndTurn);
+
```
hooperk, line 121: Hmm... more musing to myself but we should really make a conveniecne method for this standard "delete at turn end/opponent turn end" type of effect

x89rt2: Frankly, this method is still bugged last I checked. We cant get it to last multiple turns for continuous triggers. I wonder if in addition to the deletion, we can reapply itself in a continuous loop.

hooperk: I'm curious about it being an effect added to the player rather than the digimon. Maybe it's something we can take a look at in the near future

hooperk: I think the issue is it being added with an until turn end timing instead of being added forever, but I can't look into the add effects code to see if a forever option exists

CardEffect/EX11/Yellow/EX11_061.cs
x89rt2: This was copied from bt17 rika and I don't fully understanding so I left it as is. I'll take a shot, but you'll have to quadrupole check it later.

x89rt2: Alright, here's my attempt without referencing anything to see if I'm starting to learn some of this.

CardEffect/EX11/Yellow/EX11_061.cs
```diff
@@ -0,0 +1,160 @@
...
+                    return "[Your Turn] When any of your Digimon digivolve into a [Puppet] trait Digimon, by suspending this Tamer, you may play 1 level 3 [Puppet] trait Digimon card from your hand without paying the cost. At turn end, delete the Digimon this effect played.";
+                }
+
+                bool CanUseCondition(Hashtable hashtable)
+                {
+                    return CardEffectCommons.CanTriggerWhenPermanentDigivolving(permanent, PermanentCondition)
+                        && CardEffectCommons.IsExistOnBattleArea(card);
+                }
+
+                bool PermanentCondition (permamemt Permanent)
```
hooperk: Almost. Typo: Permanent permanent (The capital letter one has to be first to indicate the class)

hooperk (changes requested):

CardEffect/EX11/Yellow/EX11_061.cs
```diff
@@ -0,0 +1,160 @@
...
+                bool PermanentCondition (Permamemt permanent)
+                {
+                    return CardEffectCommons.IsPermanentExistsOnOwnerBattleAreaDigimon(permanent, card)
+                        && permanent.TopCard.EqualsTraits("Puppet");
+                }
+
+                bool CanActivateCondition(Hashtable hashtable)
+                {
+                    return CardEffectCommons.IsExistOnBattleArea(card)
+                        && CardEffectCommons.IsOwnerTurn(card)
```
hooperk: Is owner turn should be part of Use Condition

hooperk (approved):

PR #1290, opened 2026-01-18
Update BT21_059.cs

x89rt2:
Fix YT being AT
Fix YT not triggering if oppo has no digimon

### DIFF
--- a/CardEffect/BT21/Black/BT21_059.cs
+++ b/CardEffect/BT21/Black/BT21_059.cs
@@ -143,6 +143,7 @@ namespace DCGO.CardEffects.BT21
                 bool CanUseCondition(Hashtable hashtable)
                 {
                     return CardEffectCommons.IsExistOnBattleAreaDigimon(card) &&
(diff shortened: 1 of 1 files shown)


CardEffect/BT21/Black/BT21_059.cs
```diff
@@ -143,6 +143,7 @@ bool IsOpponentsDigimon(Permanent permanent)
                 bool CanUseCondition(Hashtable hashtable)
                 {
                     return CardEffectCommons.IsExistOnBattleAreaDigimon(card) &&
+                        CardEffectCommons.IsOwnerTurn(card) &&
                         CardEffectCommons.CanTriggerWhenLinked(hashtable, PermanentCondition, CardCondition) &&
                         CardEffectCommons.HasMatchConditionOpponentsPermanent(card, IsOpponentsDigimon);
```
hooperk: This ability should trigger regardless of if there is an opponent's digimon

hooperk (changes requested):

hooperk (approved):

PR #1303, opened 2026-01-19
General fix: Multiple DigiXros or Assemblies at once

hooperk:
Basic idea is rather than each of these classes representing exactly 1 digixros or assembly, they maintain a mapping between a particular card and a digixros or assembly attempt for that card, since the last time the class was reset.

The classes now also maintain a history of all cardSources already selected to be placed under cards by either class so they do not get selected again and cause conflict.

This should enable multiple digixrosses when several cards are all played at once (BT5-087 Omnimon Zwart and 2 digixros capable black and purple can test, and EX9-047 Eyesmon can test Assembly while being played alongside something else, or even 2 Eyesmon with 8 Shatter mode in trash for full test)

### DIFF
--- a/CardEffect/BT10/Purple/BT10_093.cs
+++ b/CardEffect/BT10/Purple/BT10_093.cs
@@ -114,7 +114,7 @@ namespace DCGO.CardEffects.BT10
             {
                 if (CardEffectCommons.IsExistOnBattleArea(card))
                 {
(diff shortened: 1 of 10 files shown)


x89rt2:
Do we want to merge https://github.com/DCGO2/DCGO-Card-Scripts/pull/1262 with this?

hooperk:
> Do we want to merge #1262 with this?

No, it'll be one or the other based on if we feel this one is overkill or not

PR #1308, opened 2026-01-20
Implement EX11-032 GrandGalemon

### LINKED ISSUE #770: EX11-032 - GrandGalemon
Image

hooperk:


CardEffect/EX11/Green/EX11_032.cs
```diff
@@ -0,0 +1,314 @@
...
+                    return CardEffectCommons.IsExistOnHand(card)
+                        && CardEffectCommons.IsOwnerTurn(card)
+                        && CardEffectCommons.HasMatchConditionOwnersPermanent(card, IsShoto)
+                        && CardEffectCommons.HasMatchConditionOwnersPermanent(card, IsPteromon)
+                        && CardEffectCommons.HasMatchConditionOwnersCardInTrash(card, IsGalemon);
+                }
+
+                bool IsShoto(Permanent permanent)
+                {
+                    return permanent.IsTamer && permanent.TopCard.EqualsCardName("Shoto Kazama");
```
x89rt2: no mention of shoto having to be tamer

CardEffect/EX11/Green/EX11_032.cs
```diff
@@ -0,0 +1,314 @@
...
+                }
+
+                bool IsGalemon(CardSource source)
+                {
+                    return source.IsDigimon && source.EqualsCardName("Galemon");
+                }
+
+                bool IsGrandGalemon(CardSource source)
+                {
+                    return source.IsDigimon && source.EqualsCardName("GrandGalemon");
```
x89rt2, line 53: this doesn't seem right, this would evo into any grandgale no? needs to evo specifically into this card

CardEffect/EX11/Green/EX11_032.cs
```diff
@@ -0,0 +1,314 @@
...
+                    }
+                }
+            }
+            #endregion
+
+            #region When Digivolving
+            if (timing == EffectTiming.OnEnterFieldAnyone)
+            {
+                ActivateClass activateClass = new ActivateClass();
+                activateClass.SetUpICardEffect("May Suspend 1 Digimon. May play 1 [Avian] or [Bird] withh 3k DP + 1k per suspended digimon.", CanUseCondition, card);
```
x89rt2: activateClass.SetUpICardEffect("May Suspend 1 Digimon. May play 1 [Avian] or [Bird] in traits with 3k DP + 1k per suspended digimon.", CanUseCondition, card);

CardEffect/EX11/Green/EX11_032.cs
```diff
@@ -0,0 +1,314 @@
...
+                    if (CardEffectCommons.IsExistOnBattleAreaDigimon(card))
+                    {
+                        if (CardEffectCommons.IsOwnerTurn(card))
+                        {
+                            bool WinnerCondition(Permanent permanent) => permanent.cardSources.Contains(card);
+
+                            if (CardEffectCommons.CanTriggerWhenWinBattle(
+                                hashtable: hashtable,
+                                winnerCondition: WinnerCondition,
+                                isOnlyWinnerSurvive: true))
```
x89rt2: i'll assume the code is correct but the "onlywinnersurvive" sure is misleading

x89rt2 (changes requested):

CardEffect/EX11/Green/EX11_032.cs
hooperk: We set a parameter on the Digivolve method to ensure that, this isn't really used but just in case something messed up

CardEffect/EX11/Green/EX11_032.cs
hooperk: Copied from older Liberator warp level 5s so we may need to check them, but you're right

CardEffect/EX11/Green/EX11_032.cs
hooperk: You made me realize there was still an issue with the method that this was even still included. Should more correctly only trigger for battles that have been won, and not for ties where the attacker wasn't deleted

Michael8818 (approved):

PR #1325, opened 2026-01-22
Update P_215.cs

x89rt2:
Fix to allow to grab eggs

### DIFF
--- a/CardEffect/P/Blue/P_215.cs
+++ b/CardEffect/P/Blue/P_215.cs
@@ -53,7 +53,8 @@ namespace DCGO.CardEffects.P
 
             bool CardSelectCondition(CardSource cardSource)
             {
-                return cardSource.IsDigimon &&
+                return (cardSource.IsDigimon ||
+                    cardSource.IsDigiEgg) &&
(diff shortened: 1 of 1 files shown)


CardEffect/P/Blue/P_215.cs
```diff
@@ -53,7 +53,8 @@ bool SharedCanActivateCondition(Hashtable hashtable)
 
             bool CardSelectCondition(CardSource cardSource)
             {
-                return cardSource.IsDigimon &&
+                return (cardSource.IsDigimon ||
+                    cardSource.IsDigiEgg) &&
```
hooperk: It says card, so lets instead remove the IsDigimon. If a tamer card in trash has a level somehow it is a valid target

hooperk (changes requested):

CardEffect/P/Blue/P_215.cs
x89rt2: I wasn't going to make that assumption but if you say so.

hooperk (approved):

PR #1338, opened 2026-01-22
Update CanNotUnsuspend.cs

### LINKED ISSUE #73: BT16-046 GranKuwagamon
Back to the old issue of when blast digivolved opponent's digimon unsuspend on their following turn.
https://discord.com/channels/1095717982522585098/1257400945248178218

x89rt2:
I mean, what's the worst that could happen?

In theory fixes the same bug on BT24 Ogremon as the attached issue.

### DIFF
--- a/Scripts/CardEffectCommons/GiveEffect/GiveEffectToPermanent/CanNotUnsuspend.cs
+++ b/Scripts/CardEffectCommons/GiveEffect/GiveEffectToPermanent/CanNotUnsuspend.cs
@@ -33,7 +33,7 @@ public partial class CardEffectCommons
 
         yield return ContinuousController.instance.StartCoroutine(GainCanNotUnsuspend(
             targetPermanent: targetPermanent,
-            effectDuration: EffectDuration.UntilOpponentTurnEnd,
+            effectDuration: EffectDuration.UntilNextUntap,
             activateClass: activateClass,
             condition: CanUseCondition,
             effectName: effectName


Scripts/CardEffectCommons/GiveEffect/GiveEffectToPermanent/CanNotUnsuspend.cs
```diff
@@ -33,7 +33,7 @@ bool CanUseCondition()
 
         yield return ContinuousController.instance.StartCoroutine(GainCanNotUnsuspend(
             targetPermanent: targetPermanent,
-            effectDuration: EffectDuration.UntilOpponentTurnEnd,
+            effectDuration: EffectDuration.UntilNextUntap,
             activateClass: activateClass,
             condition: CanUseCondition,
             effectName: effectName
```
hooperk, line 39: Make a second effect for this. we still want to have general cannot unsuspend for the turn effect

hooperk (changes requested):

Scripts/CardEffectCommons/GiveEffect/GiveEffectToPermanent/CanNotUnsuspend.cs
hooperk: NVM, actually looked at the class and this is a separate method. Let me just dig in a little

hooperk: I think there is one other change that needs to happen, ok if I push a commit?

x89rt2: take it from me

PR #1344, opened 2026-01-22
CEC method for checking a card was sent to security

hooperk:


x89rt2:
I'm still seeing the same general idea as the venus code being tested right now which failed. We can import this since fixing it later will be much easier since it will all be in the 1 area, but I don't believe it currently works. Could be wrong.

Specifically talking about sending token to sec is not giving a success.

hooperk:
Migrated and merged on main repo

PR #1353, opened 2026-01-23
WIP Implement ST24-07 and proposal for how dual cards can handle

hooperk:
TODO:

- [x] Arts Digivolution is not by effect. Need to separate it from activateClass.
- [ ] Change all instances of CardSource.CardColors.Contains to CardSource.HasCardColor everywhere. Not going to this effort until we confirm the approach
- [x] Make it so when player drags a DualCard that is seen as using the option
- [ ] Update json parsing to populate DualCardColors when present. and to recognise when cardkind is dualcard
- [ ] Tweaking to the right click window and deck buildign screen to display all information.
- [ ] Further changes to cEntity_Base to store the separate effect field for the option half.
- [ ] Update cards which trash Delay options such as BT19-064 Justimon, BT18-100 Gospel of the Fallen Angel to NOT delete DualCards, depending on rulings

I am not suggesting all of the above will be done by me, just trying to track all things likely to be needed

### DIFF
--- a/CardEffect/ST24/Yellow/ST24_07.cs
+++ b/CardEffect/ST24/Yellow/ST24_07.cs
@@ -1,0 +1,386 @@
+using System;
+using System.Collections;
+using System.Collections.Generic;
+
+// Shinegreymon // GeoGrey Sword
+namespace DCGO.CardEffects.ST24
+{
+    public class ST24_07 : CEntity_Effect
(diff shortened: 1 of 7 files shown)


hooperk:
Migrated to main repo https://github.com/DCGO2/DCGO/pull/624

PR #1355, opened 2026-01-23
Fix Shoutmon X7: SM not working from under Kotone

hooperk:
Change to allow a particular class of effects, labeled EffectOfCard, to always work even when under a tamer. For effects with no trigger condition such as "When this card would be played..."


hooperk:
Merged in main repo

PR #1360, opened 2026-01-24
Update EX11_066.cs

x89rt2:
Fixing AT to support multiple digimon played and adding 1 vemmon to 2. Code is needed for other cards such as EX11 Close. Might also be somewhat relevant to some older cards but have not verified yet.

### DIFF
--- a/CardEffect/EX11/Black/EX11_066.cs
+++ b/CardEffect/EX11/Black/EX11_066.cs
@@ -1,4 +1,3 @@
-using System;
 using System.Collections;
 using System.Collections.Generic;
 using System.Linq;
--- a/CardEffect/EX11/Black/EX11_066.cs
+++ b/CardEffect/EX11/Black/EX11_066.cs
@@ -263,17 +263,76 @@ namespace DCGO.CardEffects.EX11
                                 yield return ContinuousController.instance.StartCoroutine(GManager.instance.GetComponent<Effects>().ShowCardEffect2(digivolutionCards_fixed, "Digivolution Cards", true, true));
                                 yield return ContinuousController.instance.StartCoroutine(permanent.AddDigivolutionCardsBottom(digivolutionCards_fixed, activateClass)); ;
                             }
-
-                            if (selectedCards.Count > 0)
-                            {
-
-                            }                           
                         }
-                    }
 
-                    if (selectedCards.Count > 0 && CardEffectCommons.MatchConditionOwnersPermanentCount(card, permanent => targetPermanents.Contains(permanent)) >= 2)
-                    {
+                        if (selectedCards.Count > 0)
+                        {
+                            SelectPermanentEffect selectPermanentEffect1 = GManager.instance.GetComponent<SelectPermanentEffect>();
 
+                            selectPermanentEffect.SetUp(
+                                selectPlayer: card.Owner,
+                                canTargetCondition: permanent => targetPermanents.Contains(permanent),
+                                canTargetCondition_ByPreSelecetedList: null,
+                                canEndSelectCondition: null,
+                                maxCount: 1,
+                                canNoSelect: false,
+                                canEndNotMax: false,
+                                selectPermanentCoroutine: SelectPermanentCoroutine1,
+                                afterSelectPermanentCoroutine: null,
+                                mode: SelectPermanentEffect.Mode.Custom,
+                                cardEffect: activateClass);
+
+                            selectPermanentEffect1.SetUpCustomMessage("Select 1 Digimon that will get the digivolution cards.", "The opponent is selecting 1 Digimon that will get the digivolution cards.");
+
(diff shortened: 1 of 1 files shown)


x89rt2:
Hm.. with no effect before this existing that can place sources under 2 different digimon at once, I guess we'll have to do them one by one, see if first one placed 2 or 1, then if there's a vemmon remaining, choose a target again. It also needs to auto do this if there's only one target to save on clicks.

Also, since the vemmon add is mandatory, having to reveal them, click on both, end selection, then select target feels like too many clicks. We can reveal the 2 cards for both players, then have the sim auto grab the vemmon, if only one target to place under auto go, if not, select among played/digivolved digimon, choose order to place vemmon if more than 1, otherwise if 1 just auto add.

x89rt2:
I'm closer. The reveal code has player selection aspect removed to make it require less clicks. Place vemmon code still needs work. Possibly needs code similar to 

ContinuousController.instance.StartCoroutine(CardEffectCommons.SelectTrashDigivolutionCards(

for placing the vemmon under multiple digimon?

x89rt2:
A little better but not happy. If there's only 1 target to choose from, there's no reason why it has to get manually selected. It can just be auto selected. The only thing is, this doesn't yet have a method to choose the order to place the vemmon. Also, I still have nothing on if there are 2 or more played/digivolved digimon.

For the auto select, is it really just a -1 on maxCount? does it work for selectPermanentEffect too?

I guess for the 2 or more, I'll have to add a selection, choose 1 digimon, bring up a selection of the revealed vemmon, have an end before max, if max chosen, place all under in order and end code, if only 1 choosed, repeat the code if it's at all possible? otherwise add another selection similar to the last. Not a fan of using the same code twice so hope there's an easy way to code to repeat the same selection. Since there won't be a no selection option, it should not infinite loop.

x89rt2:
I think I just need a code on line 269 to jump to line 219 to have a first rough ready.

x89rt2:
well, first rough is ready

x89rt2:
Hm... technically it could be better. If there's only one digimon played/digivolve, the only choice necessary is the order to place the vemmon under. I think current rough has you manually click on the played/digivolved digimon.

x89rt2:
I feel like I'm very close to having it do exactly what I want. Just need to fix the error on line 214 by putting a code directly above it. Just need to figure out what that code is.

x89rt2:
Is that it? Does this work?

CardEffect/EX11/Black/EX11_066.cs
```diff
@@ -172,18 +172,15 @@ IEnumerator ActivateCoroutine(Hashtable hashtable)
...
-                            canTargetCondition:IsVemmon,
-                            message: "Select 2 [Vemmon] to tuck.",
-                            mode: SelectCardEffect.Mode.Custom,
-                            maxCount: 2,
-                            selectCardCoroutine: SelectCardCoroutine),
-                        },
+                                canTargetCondition: IsVemmon,
+                                message: "",
+                                mode: SelectCardEffect.Mode.Custom,
+                                maxCount: -1,
```
hooperk: -1 will make it never run, but the RevealDeckTopCardsAndProcessForAll already checks how many valid targets there are in case there are less than maxCount, it handles that

hooperk (changes requested):

CardEffect/EX11/Black/EX11_066.cs
x89rt2: so it needs a 2 here?

x89rt2:
Having issues on triggering now for some reason. Ugh.
Also the select one of 2 vemmon to place under the one guy, on second selection the first vemmon still shows....

x89rt2:
Might have it all fixed now?

x89rt2:
Close but there are 2 problems with this version.

1. Should be easy to fix but it has the No Selection option with nothing selected. Should only show on 1 or more selected.

2. Harder but on following uses, sometimes does not recognize 1 of the Vemmon and does not have it selectable.

x89rt2:
Problem 1 fixed, working on 2.

x89rt2:
Number 2 I have not been able to recreate since last fix for 1. Did that somehow fix it?

With that said, I could make it more click efficient possibly. If reveal 1 vemmon, it's having me select it and end selection....

x89rt2:
I think I have it as good as it gets unless that one bug shows up again. As a description of the bug, sometimes after a xeno is used several times, it "desynced" or something and when revealed 2 vemmon, only allowed to select 1. Has not happened for a couple of changes so I think this will be the final version outside of maybe code optimizations.

CardEffect/EX11/Black/EX11_066.cs
hooperk: Yup

hooperk (approved):

PR #1361, opened 2026-01-24
Update BT24_040.cs

x89rt2:
Prep work on most commonly accepted interpretation of Venus protection. Testing.


x89rt2:
Primary testing has it working as I would expect. The sending token to security stuff comes later separately.

hooperk (approved):

PR #1378, opened 2026-01-27
Fix EX11-027 linking to invalid targets, doesn't place to digivolution cards, and no options not to link

hooperk:
Link Cards were not correctly being cleaned from the permanent's LinkedCards array, meaning if the card was added back to the stack in any way it would still be considered a linked card

### DIFF
--- a/Scripts/CardSource.cs
+++ b/Scripts/CardSource.cs
@@ -3049,27 +3049,30 @@ public class CardSource : MonoBehaviour
         {
             if (targetPermanent.TopCard != null && !targetPermanent.TopCard.IsToken)
             {
-                if (this.CanLink(PayCost, allowBreeding))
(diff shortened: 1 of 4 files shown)


Scripts/CardSource.cs
```diff
@@ -3049,27 +3049,30 @@ public bool CanLinkToTargetPermanent(Permanent targetPermanent, bool PayCost, bo
         {
             if (targetPermanent.TopCard != null && !targetPermanent.TopCard.IsToken)
             {
-                if (this.CanLink(PayCost, allowBreeding))
+                if(allowBreeding || !targetPermanent.TopCard.Owner.GetBreedingAreaPermanents().Contains(targetPermanent))
```
hooperk, line 3052: This line is the actual change, to ensure we actually make use of allowBreeding when checking CanLinkToTarget... , as CanLink only checks if there is some valid target and will return true based on a card in the battle area even if targetPermanent is in the breeding area

Michael8818 (approved):

PR #1392, opened 2026-01-28
EX11: Fix Flip face up incorrect condition

hooperk:


Scripts/CardController.cs
```diff
@@ -5064,7 +5064,7 @@ public IFlipSecurity(CardSource source)
 
     public IEnumerator FlipFaceUp()
     {
-        if (!_player.SecurityCards.Contains(_cardSource) || _cardSource.IsFlipped)
+        if (!_player.SecurityCards.Contains(_cardSource) || !_cardSource.IsFlipped)
```
hooperk, line 5067: Had the condition backwards. Should be "If already face up, exit" but was "If already face down"

Michael8818 (approved):

PR #1395, opened 2026-01-28
Implement EX11-055 Chitose Horaiji

### LINKED ISSUE #792: EX11-055 - Chitose Horaiji
Image

hooperk:


CardEffect/EX11/Red/EX11_055.cs
```diff
@@ -0,0 +1,203 @@
...
+
+            bool ValidDiscardCard(CardSource cardSource)
+            {
+                return cardSource.EqualsTraits("Composite")
+                    || cardSource.EqualsTraits("Wicked God");
+            }
+
+            IEnumerator SharedActivateCoroutine(Hashtable hashtable, ActivateClass activateClass)
+            {
+                if (card.Owner.HandCards.Count >= 1)
```
x89rt2: I don't think we need this since canactivate already checks

CardEffect/EX11/Red/EX11_055.cs
```diff
@@ -0,0 +1,203 @@
...
+                {
+                    return CardEffectCommons.IsExistOnBattleArea(card)
+                        && CardEffectCommons.CanActivateSuspendCostEffect(card);
+                }
+
+                bool CanSelectCardCondition(CardSource cardSource)
+                {
+                    return cardSource.EqualsCardName("Gazimon")
+                        || cardSource.EqualsCardName("Gizamon");
+                }
```
x89rt2, line 142: I'm going to have to assume the canbeplayed code isn't needed here. I can't keep up with which codes need it and which don't so I won't bother.

x89rt2 (changes requested):

CardEffect/EX11/Red/EX11_055.cs
hooperk: It actually is needed here, glad you mentioned it.
For reference: As far as I can recall the CanPlayAsNewPermanent is always needed
But for digivolution, we don't need CanPlayPermanentTargetFrame if using the digivolve effects included in card effect commons,

x89rt2: I'll try to remember that then.

x89rt2 (approved):

PR #1400, opened 2026-01-29
Implement EX11-074 Vortexdramon and battle during effect

### LINKED ISSUE #811: EX11-074 - Vortexdramon
Image

hooperk:
Rather than copy the battle process, as it was already separated from the rest of the attack process I have enhanced it with an additional parameter of a card effect. If not given a card effect it is considered to be started by the attack process and acts accordingly including additional auto processing points. If card effect is passed it performs only the battle itself, although still triggers effects that care about battle.

Implemented with the following assumptions as we don't have core rules confirmation yet:
A battle does not care if there is an ongoing attack or not. It does not interact with that attack at all, and can be performed even if an attack is ongoing.
Starting a battle does not require suspending nor being unsuspended. It may occur during the opponent's turn.
All effects that care about battle such as cannot be deleted by battle, would be deleted by battle and was deleted by battle still occur as per a battle.
You may start a battle by effect even vs a Digimon with immunity to card effects/your Digimon effects. (this one somewhat confirmed by the reveal stream)
Piercing checks are not performed due to this not being part of an attack.

### DIFF
--- a/CardEffect/EX11/Green/EX11_074.cs
+++ b/CardEffect/EX11/Green/EX11_074.cs
@@ -1,0 +1,291 @@
+using System;
+using System.Collections;
+using System.Collections.Generic;
+using System.Linq;
+
+// Vortexdramon
+namespace DCGO.CardEffects.EX11
+{
+    public class EX11_074 : CEntity_Effect
+    {
+        public override List<ICardEffect> CardEffects(EffectTiming timing, CardSource card)
+        {
+            List<ICardEffect> cardEffects = new List<ICardEffect>();
+
+            #region Static Effects
+
+            #region Alternate Digivolution
+            if (timing == EffectTiming.None)
+            {
+                bool Condition()
+                {
+                    return CardEffectCommons.HasMatchConditionOwnersPermanent(card, HasShotoKazama);
+                }
+
+                bool HasShotoKazama(Permanent targetPermanent)
+                {
+                    return CardEffectCommons.IsPermanentExistsOnOwnerBattleArea(targetPermanent, card) &&
+                           targetPermanent.TopCard.EqualsCardName("Shoto Kazama");
+                }
+
+                bool PermanentCondition(Permanent targetPermanent)
+                {
+                    return CardEffectCommons.IsPermanentExistsOnOwnerBattleAreaDigimon(targetPermanent, card)
+                        && targetPermanent.TopCard.EqualsCardName("GrandGalemon");
+                }
+
+                cardEffects.Add(CardEffectFactory.AddSelfDigivolutionRequirementStaticEffect(
+                    permanentCondition: PermanentCondition,
+                    digivolutionCost: 6,
+                    ignoreDigivolutionRequirement: false,
+                    card: card,
+                    condition: Condition)
+                );
+            }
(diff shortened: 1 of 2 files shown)


Scripts/CardController.cs
```diff
@@ -4447,37 +4453,44 @@ bool ShowEffect()
 
                 #endregion
 
-                #region effect when determine whether to do security check
+                if (CardEffect == null)//If started by effect, do not perform piercing
```
hooperk: Technically from a rules perspective, Piercing is supposed to trigger and activate, but fail to do anything.
From our code perspective: nothing cares about seeing a Piercing effect activate, and if we activate it anyway it will probably mess up/interact with the ongoing attack somehow. So for our sanity, just not even triggering it

x89rt2:
With the JP QNA being out, we need to make sure this supports attacking vortexdra piercing triggering from the effect battle but not normal battle and still doing the sec check.

CardEffect/EX11/Green/EX11_074.cs
```diff
@@ -0,0 +1,308 @@
...
+                            {
+                                if (cardSource == card)
+                                {
+                                    return true;
+                                }
+                            }
+                        }
+
+                        return false;
+                    }
```
x89rt2, line 138: This could be cleaned up. The below one too I guess.

x89rt2 (dismissed):

x89rt2:
Special note, this card's AT will need the hash refresh code whenever that one is ready. It might be worth altering code to prep for it. Just a check that if the AT both no unsuspend and no battle are selected, have an empty set for refresh code.

hooperk:
> With the JP QNA being out, we need to make sure this supports attacking vortexdra piercing triggering from the effect battle but not normal battle and still doing the sec check.

I did double check all the code on Piercing and am confident it will work correctly for that. I have left it so it does trigger and activate during a battle by effect. But the CanActivateCondition of Piercing itself will ensure it only activates if it is also the attacking digimon of the ongoing attack process (if there is one). When it does successfully activate, it will set the flag on the attack process to allow piercing checks. This flag is only reset at the very end of an attack and the very start, during a successful attack declaration, which must have been already completed for an attack to be ongoing.

x89rt2:
Vortexdra is approved. I'll leave it to mike for the cardcontroller.

Scripts/CardController.cs
Michael8818, line 1: The problem i have with this is what if the fix for all the attack by effects timing things might involving passing an effect here? this isn't very specific to the scenario going on. Would rather just a bool be passed if it was triggered by an attack or not wouldn't that make more sense?

Michael8818 (changes requested):

Scripts/CardController.cs
hooperk: True. I passed effect thinking I'd need to reference it and then it never came up

Michael8818, line 1: this isn't exactly what I meant. lets rename that bool IsWithoutAttack so it can specifically be used for this instance of battling without any attack.

Michael8818 (changes requested):

Michael8818 (approved):

PR #1422, opened 2026-02-05
Update SelectDigiXrosClass.cs

x89rt2:
Change on line 385

//Previously, the break was commented out in order to resolve the below problem. I'm reintroducing the break to diagnose what digixross were not triggering.
//break; Removed for not triggering digixros in all situations


hooperk:
This looks like a post EX11 test, right?

x89rt2:
> This looks like a post EX11 test, right?

correct

PR #1428, opened 2026-02-05
Update EX11_011.cs

x89rt2:
Fix a problem with enemy target selection on shared effect. Not convinced fixes reported issue:

https://discord.com/channels/1095717982522585098/1468822845537648650


hooperk (approved):
Pretty sure this fixes exactly that softlock. ArrayOutOfBoundException trying to get the first item of an empty array

PR #1431, opened 2026-02-06
Fix EX11-059 All Turns not digivolving, Fix EX11-059, EX11-070 can choose trash card if no valid hand card

hooperk:
I'll be honest, this isn't a full fix, but the local one has too many whited out fields for me to see what is going on so I need to push what I have for now and get video


Scripts/CardEffectCommons/DNADigivolveEffects.cs
```diff
@@ -313,7 +313,7 @@ IEnumerator SelectDNACardCoroutine(CardSource source)
             yield return null;
         }
 
-        if (isWithHandCard && owner.HandCards.Some(cardSource => CanJogressWithHandOrTrash(cardSource, owner, isWithHandCard, isIntoHandCard, targetCardCondition, permanentCondition, digivolutionCardCondition)))
+        if (isIntoHandCard && owner.HandCards.Some(cardSource => CanJogressWithHandOrTrash(cardSource, owner, isWithHandCard, isIntoHandCard, targetCardCondition, permanentCondition, digivolutionCardCondition)))
```
x89rt2: At work so cant check. 

CanJogressWithHandOrTrash(cardSource, owner, isWithHandCard,

Still has iswithhandcard. Is that fine?

hooperk: Yes, still uses both. IsWithHand is whether the cardsource is a hand or trash card, isIntoHand is whether the target to dna into is from the hand or trash. This function allows you to DNA with 1 permanent in play and 1 card in hand or trash into another card in hand or trash

Michael8818 (approved):

PR #1432, opened 2026-02-06
Update EX11_053.cs

x89rt2:
Fix some parenthesis for OP. We'll see what happens.

Incorrect int on OD.


CardEffect/EX11/White/EX11_053.cs
```diff
@@ -66,8 +66,8 @@ bool CanActivateCondition(Hashtable hashtable)
                 bool CanSelectPermanentCondition(Permanent permanent)
                 {
                     return (CardEffectCommons.IsPermanentExistsOnOwnerBattleAreaDigimon(permanent, card)
```
hooperk: Remove Digimon from this check then so we aren't checking that twice

hooperk (changes requested):

hooperk (approved):

PR #1433, opened 2026-02-06
Update EX11_071.cs

x89rt2:
First suspicious thing I see in code, but I don't know if this code takes into account egg drasil discount.

### DIFF
--- a/CardEffect/EX11/White/EX11_071.cs
+++ b/CardEffect/EX11/White/EX11_071.cs
@@ -195,7 +195,7 @@ namespace DCGO.CardEffects.EX11
                                         || cardSource.EqualsTraits("LIBERATOR"));
                             }
 
-                            bool RootCondition(SelectCardEffect.Root root) =>true;
+                            bool RootCondition(SelectCardEffect.Root root) => true;
 
                             bool isUpDown() => true;
 
--- a/CardEffect/EX11/White/EX11_071.cs
+++ b/CardEffect/EX11/White/EX11_071.cs
@@ -102,7 +102,7 @@ namespace DCGO.CardEffects.EX11
                     return (cardSource.EqualsTraits("Royal Knight")
                         || cardSource.EqualsTraits("Liberator"))
                         && cardSource.GetCostItself >=4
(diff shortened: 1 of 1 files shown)


CardEffect/EX11/White/EX11_071.cs
```diff
@@ -100,7 +100,7 @@ bool CanUseCondition(Hashtable hashtable)
                 bool CanSelectCardCondition(CardSource cardSource)
                 {
                     return (cardSource.EqualsTraits("Royal Knight")
-                        || cardSource.EqualsTraits("Liberator"))
+                        || cardSource.EqualsTraits("LIBERATOR"))
```
hooperk: EqualsTraits ignores case so there is likely some other issue as well

x89rt2: Did that get fixed at some point cause last I checked, it did not ignore case.

x89rt2:
I've done changes, but none of them are really impactful.

x89rt2:
Had to remove 

&& cardSource.GetCostItself >= 4

on cost reduction code to fix stack overflow. Not sure if that introduces any new problems/illegal plays but I guess someone will figure it out.

CardEffect/EX11/White/EX11_071.cs
```diff
@@ -166,77 +166,37 @@ ICardEffect GetCardEffect(EffectTiming _timing)
...
+                                        || targetPermanents.Count((targetPermanent) => targetPermanent != null) == 0;
                             }
 
                             bool CardSourceCondition(CardSource cardSource)
                             {
-                                if (cardSource != null)
-                                {
-                                    if (cardSource.Owner == card.Owner)
-                                    {
-                                        if (cardSource.GetCostItself >= 4)
```
hooperk, line 213: Yes this needed to be removed since this Get CostItself is in turn checking the conditions of all activ cost reductions which means this would then check again and again...
Should be fine since the main selection already chose a valid Digimon

hooperk (approved):

PR #1434, opened 2026-02-07
Update EX11_066.cs

x89rt2:
Let's see if this fixes the weird Xeno issue or just fully breaks the effect.

### DIFF
--- a/CardEffect/EX11/Black/EX11_066.cs
+++ b/CardEffect/EX11/Black/EX11_066.cs
@@ -182,7 +182,7 @@ namespace DCGO.CardEffects.EX11
                                     canTargetCondition: cardSource => cardSource.EqualsCardName("Vemmon"),
                                     message: "",
                                     mode: SelectCardEffect.Mode.Custom,
-                                    maxCount: 2,
+                                    maxCount: -1,
                                     selectCardCoroutine: SelectCardCoroutine),
                             remainingCardsPlace: RemainingCardsPlace.Trash,
                             activateClass: activateClass


CardEffect/EX11/Black/EX11_066.cs
```diff
@@ -182,7 +182,7 @@ IEnumerator ActivateCoroutine(Hashtable hashtable)
                                     canTargetCondition: cardSource => cardSource.EqualsCardName("Vemmon"),
                                     message: "",
                                     mode: SelectCardEffect.Mode.Custom,
-                                    maxCount: 2,
+                                    maxCount: -1,
```
hooperk, line 185: When maxCount is 0 or less no selection is done at all

hooperk (changes requested):

CardEffect/EX11/Black/EX11_066.cs
x89rt2: I've done some primary testing myself in open and have not encountered this issue. I can properly add vemmon under guys.

x89rt2: for the record, this isn't normal selection, this is RevealDeckTopCardsAndProcessForAll which normally just shows stuff on -1 and then places them back in deck or trashes or something which I've bastardized to work like this.

hooperk: OK, if you've tested it I'll take your word for it

hooperk (approved):

PR #1443, opened 2026-02-08
Post EX11: Fix BT24-090 Performance

hooperk:
Got my approach typed out but we can try Lewis's first.
Basically applies the alliance effect onto every relevant card on the board while this card is in security, and allows the CanUseCondition of that effect to track whether or not it still applies.
This should be more efficient than adding the effect every single effect check and then also checking the conditions.
Applies to every single card source to ensure it still works through any combination of dedigivolving or evolving or source tucking etc., but then only the topcard applies.
Uses Card Index to track if effect has already been added, in such a way that if the user ever somehow gets 2 in security they will correctly have 2 instances of Alliance from them. (Use Throne, Use Amethyst Mandala, Use Second identical Throne for example)
Applies to all cards upon itself being placed in security, then applies whenever a new card enters either as a new permanent entirely, a digivolution or if added to an existing permanent. Only ever for your side of the board though.

Same approach can work for Scapegoat, but will need to update Permanent.HasScapeGoat to actually check cardEffect.CanTrigger(null), as otherwise every one of your Digimon will show as having scapegoat under this approach

### DIFF
--- a/CardEffect/BT24/Blue/BT24_090.cs
+++ b/CardEffect/BT24/Blue/BT24_090.cs
@@ -36,6 +36,7 @@ namespace DCGO.CardEffects.BT24
 
             #region All Turns - Security
 
+            #region Blocker
             if (timing == EffectTiming.None)
             {
                 bool PermanentCondition(Permanent permanent)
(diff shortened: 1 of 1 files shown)


x89rt2:
Will this show as tamers having alliance in right click menu? I know we had that issue with blitz takuya at some point.

hooperk:
> Will this show as tamers having alliance in right click menu? I know we had that issue with blitz takuya at some point.

No, tamers are not actually given the effect, although such permanents are checked in case any Digimon cards are in their stack. Only cards with DP are actually given the effect (Digimon and Digieggs that can exist in battle area) but then further permanent.HasAlliance checks the CanUse condition of the alliance to ensure it actually has it, so will show in the correct scenarios for this approach.

hooperk:
Migrated to main repo https://github.com/DCGO2/DCGO/pull/625

PR #1445, opened 2026-02-08
Update P_094.cs

x89rt2:
Modernized the OP/WD, remove some unnecessary code, and remove a click in certain situations. Still working on finding a way to add an int value to a description that updates between effect activations but I'm not even sure if that's possible.

I think it's not possible without changing core code so I'm dropping that aspect.

### DIFF
--- a/CardEffect/P/Black/P_094.cs
+++ b/CardEffect/P/Black/P_094.cs
@@ -97,7 +97,7 @@ public class P_094 : CEntity_Effect
             return CardEffectCommons.IsExistOnBattleArea(card)
                 && card.Owner.Enemy.GetBattleAreaPermanents().Count(SharedCanSelectPermanentCondition) >= 1;
         }
-      
+
         bool SharedCanSelectPermanentCondition(Permanent permanent)
         {
             return CardEffectCommons.IsPermanentExistsOnOpponentBattleArea(permanent, card)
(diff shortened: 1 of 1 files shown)


x89rt2:
Oh, I could probably improve the digixros

x89rt2:
Removed 1 unnecessary click. It looks like the true I removed from the digixrossconditionelement was supposed to end all future selections if you click end selection? It doesn't actually work. Is that part of the shout superior x7 bug? Anyways, removing the true removed extra clicks in a situation where vemmon where only in hand.

CardEffect/P/Black/P_094.cs
```diff
@@ -1,281 +1,239 @@
...
+                    canEndSelectCondition: null,
+                    maxCount: 1,
+                    canNoSelect: false,
+                    canEndNotMax: false,
+                    selectPermanentCoroutine: null,
+                    afterSelectPermanentCoroutine: null,
+                    mode: SelectPermanentEffect.Mode.Destroy,
+                    cardEffect: activateClass);
+
+                yield return ContinuousController.instance.StartCoroutine(selectPermanentEffect.Activate());
```
hooperk: Why select at all? Why not just selectedPermanent = card.Owner.Enemy.GetBattleAreaPermanents().Filter(SharedCanSelectPermanentCondition)[0]

x89rt2: Cause I forgot that command existed.

hooperk (approved):

PR #1453, opened 2026-02-11
Update EX11_022.cs

x89rt2:
Add delete at turn end.

### DIFF
--- a/CardEffect/EX11/Yellow/EX11_022.cs
+++ b/CardEffect/EX11/Yellow/EX11_022.cs
@@ -156,6 +156,35 @@ namespace DCGO.CardEffects.EX11
                         if (selectedCard != null) yield return ContinuousController.instance.StartCoroutine(
                             CardEffectCommons.PlayPermanentCards(new List<CardSource>() { selectedCard }, activateClass, false, false, SelectCardEffect.Root.Trash, true));
                     }
+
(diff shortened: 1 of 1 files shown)


CardEffect/EX11/Yellow/EX11_022.cs
```diff
@@ -156,6 +156,35 @@ IEnumerator SelectCardCoroutine(CardSource cardSource)
...
+                    CardEffectCommons.AddEffectToPlayer(effectDuration: EffectDuration.UntilEachTurnEnd, card: card, cardEffect: activateClass1, timing: EffectTiming.OnEndTurn);
+
+                    bool CanUseCondition2(Hashtable hashtable)
+                    {
+                        return true;
+                    }
+
+                    bool CanActivateCondition1(Hashtable hashtable)
+                    {
+                        return selectedPermanent.TopCard != null
```
hooperk, line 177: Wouldn't this need selectedPermanent != null as well?

hooperk: NVM, the permanent does in fact exist at the time selectedPermanent is made, checking TopCard is how we make sure it is still a digimon on the board when this goes to resolve, all good

hooperk (approved):

PR #1454, opened 2026-02-11
Update EX11_044.cs

x89rt2:
I'm not sure what this was doing or if it fixes the reported problem?
image


hooperk (approved):
Yeah that actually looks like it would only let you finish selecting if you had not selected anything... lets remove that.

PR #1461, opened 2026-02-12
Update EX9_067.cs

x89rt2:
YT triggers when played out by a digimon with a wd of play a tamer.

### DIFF
--- a/CardEffect/EX9/Yellow/EX9_067.cs
+++ b/CardEffect/EX9/Yellow/EX9_067.cs
@@ -79,7 +79,8 @@ namespace DCGO.CardEffects.EX9
 
                 bool CanUseCondition(Hashtable hashtable)
                 {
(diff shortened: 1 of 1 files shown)


x89rt2:
This formatting is a little weird, not sure this fix works.

CardEffect/EX9/Yellow/EX9_067.cs
```diff
@@ -97,7 +98,6 @@ bool CanActivateCondition(Hashtable hashtable)
                     return CardEffectCommons.IsExistOnBattleArea(card) && CardEffectCommons.IsOwnerTurn(card);
```
hooperk, line 98: IsOwnerTurn should be in CanUse

CardEffect/EX9/Yellow/EX9_067.cs
hooperk, line 87: Consider collapsing this into return X && y as well

hooperk (approved):

